PR #11, opened 2018-02-23
changed :use for :request. Fixes #7

### LINKED ISSUE #7: Unclear dependencies
I am becoming even more convinced by the minute that importing dependencies through `(:use` in the namespace is an idea spawned by the devil.

It makes the following code "unreadable" in the sense that you have to know by heart whether the function comes from that namespace or is imported from another.

My proposal is to modify all uses of:

```clojure
(ns card-game.whatever
  (:use card-game.potato
        some-library.module.lolcat))
```

to:

```clojure
(ns card-game.whatever
  (:require [card-game.potato :as potato]
            [some-library.module.lolcat :as lolcat]))
```

so that we reference `(potato/random-function` instead of `(random-function`.

If the module name is too long and annoying, shorten it but keep it readable:

```clojure
(ns card-game.whatever
  (:require [card-game.potato :as potato]
            [some-library.module.breakfast-at-mcdonalds :as breakfast]))
```

Masclins:


backend/src/card_game/api.clj
```diff
@@ -1,7 +1,7 @@
 (ns card-game.api
-  (:use card-game.core
-        card-game.victory-conditions
-        card-game.persistence))
+  (:require [card-game.core :as core]
+            [card-game.victory-conditions :as victory-conditions]
+            [card-game.persistence :as persistence))
```
kenan-rhoton: missing a `]` before `))`

PR #12, opened 2018-02-25
Frontend

kenan-rhoton:
The game is now "playable" (and very very ugly).

By doing `docker-compose up --build` you can play it locally buy going to `localhost:8000` in your browser.

There is still a little problem still with Travis, but while I solve it you can do your code review.

Also, I'll put a LOT of issues on tonight (7 written down right now), so hang on tight and let's get to work!


frontend/src/components/JoinGame.vue
```diff
@@ -0,0 +1,27 @@
...
+</template>
+
+<script>
+import axios from "axios";
+export default {
+  name: "JoinGame",
+  created: async function() {
+    try {
+      const gameID = this.$route.params.joinID;
+      const response = await axios.post("http://backend:3000/games/" + gameID);
```
Masclins: I assume somehow this calls core.api/add-player
Would it be possible to comment that so it's more understandable for us plebeians?

backend/src/card_game/api_handler.clj
```diff
@@ -5,24 +5,45 @@
...
+
+(defn parse-int [number-string]
+  (try (Integer/parseInt number-string)
+       (catch Exception e nil)))
 
 (defn play-action
-  [game player action]
-  (api/play-card-as-player game player (:index action) (:row action))
-  )
+  [game-id player-id action]
```
Masclins: Thanks for using `game-id` and `player-id`

e2e/game_page.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+import { Selector } from 'testcafe'; // first import testcafe selectors
+
+export default class GamePage {
+    constructor () {
+        this.hand = Selector('.hand');
+        this.cardsInHand = this.hand.find('.card');
+        this.joinLink = Selector('#join-link');
+        this.rows = Selector('.game-row');
+        this.myScore = Selector('#my-score');
```
Masclins: `myScore` and `opponentScore` is how many rows are won?
Given that we have 6 scores (points on each row and number of rows won) we should try to be more clear regarding that.

e2e/main_page.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+import { Selector } from 'testcafe'; // first import testcafe selectors
+
+export default class MainPage {
+    constructor () {
+        this.createGame = Selector('#create-game');
```
Masclins: Where can I find `Selector('#create-game')`?

e2e/test.js
```diff
@@ -0,0 +1,90 @@
+"use strict";
+import { Selector, ClientFunction } from 'testcafe';
+import MainPage from './main_page.js';
+import GamePage from './game_page.js';
+
```
Masclins: Not reviewing (yet) since I don't understand JavaScript tests :(

frontend/src/components/CreateGame.vue
```diff
@@ -0,0 +1,27 @@
...
+    </div>
+</template>
+
+<script>
+import axios from "axios";
+export default {
+  name: "CreateGame",
+  beforeRouteEnter: async function(to, from, next) {
+    try {
+      const response = await axios.post("http://backend:3000/games");
```
Masclins: I assume somehow this calls core.api/add-player
Would it be possible to comment that so it's more understandable for us plebeians?

e2e/game_page.js
kenan-rhoton: Currently they don't work (there's no test that actually tests them) but that would be the idea.

e2e/main_page.js
kenan-rhoton: The actual element is in frontend/src/views/Main.vue. The `#` means it's an ID selector (check out [CSS Selectors](https://www.w3schools.com/CSSref/css_selectors.asp) for more info

e2e/test.js
kenan-rhoton: These aren't per se the Javascript tests (those are in frontend/test/unit/), this are the End-to-end tests, which attempt to behave like a normal user and check that things "work".

frontend/src/components/CreateGame.vue
kenan-rhoton: The interface is defined in card-game.api-handler (inside the `(defroutes app-routes` object)

PR #27, opened 2018-02-26
Potatoes

ruchnar:


backend/test/card_game/api_test.clj
```diff
@@ -2,6 +2,7 @@
   (:require [expectations.clojure.test :refer :all]
             [card-game.api :as api]))
 
+
```
Masclins: Clearly this line makes tests fail...

PR #30, opened 2018-02-27
Usable frontend

### LINKED ISSUE #13: Frontend needs urgent Styling to be useable
Frontend "works" but cannot reasonably be used by a human.

This would be fixed by:

- Applying CSS styles to frontend/src/views and components or,
- More likely: importing [Bootstrap](https://getbootstrap.com) in Frontend/public/index.html and using Bootstrap classes to make everything nice in the components and views.

kenan-rhoton:
This should be enough to close #13, it's not particularly *pretty* but it's definitely human-useable.


frontend/src/components/Game.vue
```diff
@@ -1,36 +1,44 @@
...
-                :rownum="rownum"
-                v-on:dragover="allowDrop"
-                v-on:drop="dropOnRow">
-                <div v-for="card in row"
-                    :key="card.$index"
+        <div v-if="response.status != 'Waiting for an opponent'"
+            id="board"
+            class="row">
+            <div id="my-score" class="col text-right">0</div>
+            <div id="opponent-score" class="col text-left">0</div>
```
Masclins: We still haven't defined what "my score" and "opponent score" means. Number of rows won at the moment? It's necessary to specify that for whoever implements it.

kenan-rhoton: Yes, and that is already another issue, this is just about styles

PR #34, opened 2018-02-28
Split game-board into my cards and opponent's

### LINKED ISSUE #15: Clarify game board
Change the play-card behavior so that it more clearly delimits which card is who's instead of the very ugly "Owner" tag.

One options is to divide the play area into TWO areas (my-area and opponent-area) and make cards go there. Bonus points for symmetry.

kenan-rhoton:
This is meant to close #15.

It makes the game board be split into two sections: my cards and my opponent's.

It is very clear but it could be prettier, we really need someone to work on the aesthetics at some point.


e2e/test.js
```diff
@@ -78,10 +78,10 @@ test('Sample Game', async t=> {
                 gamePage.cardsInHand.nth(0),
                 gamePage.rows.nth(0))
             .expect(gamePage.cardsInHand.count).eql(num-1)
-            .expect(gamePage.rows.nth(0).find('.card').count).eql(cardsInPlay+2)
+            .expect(gamePage.rows.nth(0).find('.card').count).eql(cardsInPlay+1)
```
Masclins: Aren't we now missing a test to see the opponent's row has the expected number of cards?

kenan-rhoton: True, also, I just realized that it CAN be unit-tested 🤔

kenan-rhoton:
That is not a review... please don't accept changes without a review, it's ok if you don't want to review it, though, just don't accept then 😃

e2e/test.js
pernilsalat: gamePage.rows.nth(0) should be declared outside the loop (it appears too many times), the rest could also recive the same treatment, but it's way less dramatic, and num variable name could also be renamed to maxHandNum or something like that.

After the second dragToElement i would add a
`
.then(() => cardsInPlay+=1)
`
this way the variable cardsInPlay is actually storing the number of cards in play

kenan-rhoton: putting that `.then` in would imply needing a second `await t`... and I don't think it woud be either more readably or more correct, it's just something we want to update every loop.

Agreed on the rest.

e2e/test.js
```diff
@@ -50,40 +50,42 @@ test('Game links are different', async t => {
...
         .expect(gamePage.cardsInHand.exists).ok()
     const opponentURL  = await getWindowLocation();
+    
     await t
         .navigateTo(myGameURL.href)
         .expect(gamePage.cardsInHand.exists).ok();
+    
     var cardsInPlay = 0;
-    var num = 12;
+    var maxHandNum = await gamePage.cardsInHand.count;
```
Masclins: Great!
That will make changing the number of cards in hand really easy (in case we want that at some point)

e2e/test.js
```diff
@@ -50,40 +50,42 @@ test('Game links are different', async t => {
 });
 
 test('Sample Game', async t=> {
```
Masclins: This mean that `test` is called as `t`?
Why? Doesn't it make it less readable? (especially on *e2e/game_page.js*)

kenan-rhoton: Yes it does, that's copy paste from the testcafe examples, will change!

e2e/test.js
```diff
@@ -20,71 +20,73 @@ test('Create a game', async t => {
...
             .dragToElement(
                 gamePage.cardsInHand.nth(0),
                 gamePage.rows.nth(0))
-            //.expect(gamePage.cardsInHand.count).eql(num-1)
-            //.expect(gamePage.rows.nth(0).find('.card').count).eql(cardsInPlay+1)
             .navigateTo(opponentURL.href)
-            .expect(gamePage.rows.nth(0).find('.card').count).eql(cardsInPlay)
-            .expect(gamePage.cardsInHand.count).eql(num)
+
+        await gamePage.checkState(testCase, maxHandNum, cardsInPlay)
```
Masclins: Clearer to merge all the tests here.

e2e/test.js
```diff
@@ -20,71 +20,73 @@ test('Create a game', async t => {
...
-    await t
+    await testCase
 	.expect(link2).notEql(link1);
 
 });
 
-test('Sample Game', async t=> {
-    await t
-        .click(mainPage.createGame)
+test('Sample Game', async testCase => {
```
Masclins: Thanks for changing the names

PR #36, opened 2018-02-28
Clean backend to close #26

### LINKED ISSUE #26: General backend structure cleanup
We need to clean up the backend project structure.

In particular I'm worried about:

- *API*: I think we should split `api.clj` into two files (with one of them only containing the actions directly reference in `api_handler.clj` and the other containing the "helper" functions. In any case, I think they should be in a subdirectory and all be contained within a `card-game.api.*` namespace.
- *Tests*: Seriously, `api-test` and `core-test` are both over 150 lines. Also the `(defexpect' groupings are kinda random. We need to restructure the tests to be more clearly focused on different functionalities by file. Also, "API" is not a functionality, it's just an interface...
- *Core*: I think we should also split up this file but I'm unsure how.
- *Build Card*: We should burn this with fire. Both the tests and its functionality and replace it with whatever it does. I don't know what I was smoking to think it was necessary for some reason.

Masclins:
This closes #26 

As commented:

- Splits *api.clj* into *api/base.clj* and *api/helper.clj*
- Creates *configs.clj* to follow DRY philosophy (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)
- (almost) Removes *build_card* (Docker don't let me remove *build_card_test.clj*)

As a bonus, the new values for cards, wanted for realease v0.2.0 is done.


Masclins:
Seems `configs/ini-hand` needs to have 12 cards to pass frontend tests.
I don't understand those test enough to change them.

backend/test/card_game/build_card_test.clj
```diff
@@ -1,13 +1,3 @@
-(ns card-game.build-card-test
-  (:require [expectations.clojure.test :refer :all]
-            [card-game.build-card :as build-card]))
+(ns card-game.build-card-test)
```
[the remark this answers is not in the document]
Masclins: I couldn't find any file referencing this namespace nor the file. Yet Docker seems to require it...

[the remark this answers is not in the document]
kenan-rhoton: 1. `docker-compose down`
2. remove the file
3. `docker-compose run backend lein clean`
4. `docker-compose run backend lein test` <- should no longer give errors

backend/src/card_game/api/helper.clj
```diff
@@ -0,0 +1,47 @@
...
+            [card-game.victory-conditions :as victory-conditions]
+            [card-game.persistence :as persistence]
+            [configs :as configs]))
+
+(defn player-num
+  "Translates a Player id into an internal player representation"
+  [game-state id] (.indexOf (:player-ids game-state) id))
+
+(defn translate-player
+  "Translates an internal player to a useful representation"
```
kenan-rhoton: "to a human-readable representation" would be more accurate, I think, might as well change these things while we're at it

backend/src/card_game/api/helper.clj
```diff
@@ -0,0 +1,47 @@
...
+  [game-state internal me]
+  (cond
+    (nil? internal) nil
+    (>= internal 2) :tie
+    (= me (get-in game-state [:player-ids internal])) :me
+    :else :opponent))
+
+(defn get-game-as-player
+  "Returns the part that ought to be visible to the player"
```
kenan-rhoton: "the part *of the game-state*"

backend/src/card_game/api/helper.clj
```diff
@@ -0,0 +1,47 @@
...
+   :hand (get-in game-state [:players (player-num game-state player-id) :hand])
+   :rows (mapv #(mapv (fn [card]
+                        (assoc card :owner
+                          (translate-player game-state (:owner card) player-id)))
+                      %)
+               (:rows game-state))
+   :winner (translate-player game-state (victory-conditions/winner game-state) player-id)})
+
+(defn define-status
+  "Returns the status of the game"
```
kenan-rhoton: "... from a player's perspective"

backend/src/card_game/api/base.clj
```diff
@@ -0,0 +1,40 @@
...
+            (persistence/save-game (core/play-card game-state (helper/player-num game-state player) index row))
+            (get-game game-id player))
+          {:error configs/out-of-turn})))
+
+(defn add-player
+  "Adds a player to a game"
+  [game-id]
+  (let [saved-game (persistence/fetch-game game-id)
+        players-connected (count (:player-ids saved-game))]
+    (if (> players-connected 1)
```
kenan-rhoton: I just realized a way to simplify this function would be to move the check for "too many players" onto `helper/create-player` (or am I missing something?)

backend/src/card_game/api/helper.clj
```diff
@@ -0,0 +1,47 @@
...
+  [game-state player-id]
+  (cond (= (count (:player-ids game-state)) 1) configs/no-opp
+        (nil? (get-in game-state [:next-play (player-num game-state player-id)])) configs/play
+        :else configs/wait))
+
+(defn uuid [] (str (java.util.UUID/randomUUID)))
+
+(defn create-player
```
kenan-rhoton: We could put the check for "too many players" here, I think

backend/src/card_game/build_card.clj
```diff
@@ -1,8 +0,0 @@
-(ns card-game.build-card
```
kenan-rhoton: 🙌

backend/src/card_game/core.clj
```diff
@@ -9,18 +9,16 @@
 
 (defn new-player
   "Creates a new player object"
-  ([] (new-player 12))
-  ([cards]
+  ([]
```
kenan-rhoton: Did we lose the ability to create players with different starting hands? (not that it was ever an extremely useful ability or one that we actually needed)

backend/test/card_game/api_messages_test.clj
```diff
@@ -0,0 +1,63 @@
...
+        :game-id
+        (api/add-player)
+        :status))
+  (expect
+    configs/no-opp
+    (let [game (api/create-game)]
+      (:status (api/get-game (:game-id game) (:player-id game)))))
+  (expect
+    configs/wait
+    (let [game (api/create-game)]
```
kenan-rhoton: I think this test can be more readable by adding `game-id (:game-id game)` to the `let` and changing `(->` syntax for a `(do` syntax.

backend/test/card_game/api_messages_test.clj
```diff
@@ -0,0 +1,63 @@
...
+      (-> (:game-id game)
+          (api/add-player)
+          :game-id
+          (api/play-card-as-player (:player-id game) 0 0)
+          :status)))
+  (expect
+    configs/play
+    (let [game (api/create-game)
```
kenan-rhoton: same as above

backend/test/card_game/api_messages_test.clj
```diff
@@ -0,0 +1,63 @@
...
+          opponent-id (:player-id (api/add-player (:game-id game)))]
+      (-> (:game-id game)
+          (api/play-card-as-player (:player-id game) 0 0)
+          :game-id
+          (api/play-card-as-player opponent-id 0 0)
+          :status)))
+  (expect
+    configs/play
+    (let [game (api/create-game)
```
kenan-rhoton: this one too

backend/test/card_game/api_messages_test.clj
```diff
@@ -0,0 +1,63 @@
...
+          (api/play-card-as-player opponent-id 0 0)
+          :game-id
+          (api/get-game (:player-id game))
+          :status))))
+
+(defexpect out-of-turn
+  ; Game gives error when playing and shouldn't
+  (expect
+    {:error configs/out-of-turn}
+    (let [game (api/create-game)]
```
kenan-rhoton: again, game-id in let

backend/test/card_game/api_messages_test.clj
```diff
@@ -0,0 +1,63 @@
...
+      (-> (:game-id game)
+          (api/add-player)
+          :game-id
+          (api/play-card-as-player (:player-id game) 0 0)
+          :game-id
+          (api/play-card-as-player (:player-id game) 0 0))))
+  (expect
+    {:error configs/out-of-turn}
+    (let [game (api/create-game)]
```
kenan-rhoton: I think this one will also be slightly more readable with that change

backend/src/configs.clj
```diff
@@ -0,0 +1,27 @@
...
+
+(def out-of-turn "Out of turn play")
+
+(def no-opp "Waiting for an opponent")
+
+(def play "Playing")
+
+(def wait "Waiting for opponent's play")
+
+(def too-players "Too many players")
```
kenan-rhoton: please name this `too-many-players`, it's much more readable

backend/src/configs.clj
```diff
@@ -0,0 +1,27 @@
+(ns configs)
+
+(def out-of-turn "Out of turn play")
+
+(def no-opp "Waiting for an opponent")
+
+(def play "Playing")
```
kenan-rhoton: Please name this `playing`, play is an action verb 😱

backend/src/configs.clj
```diff
@@ -0,0 +1,27 @@
...
+
+(def wait "Waiting for opponent's play")
```
kenan-rhoton: same here, `waiting`

backend/src/configs.clj
```diff
@@ -0,0 +1,27 @@
...
+
+(def too-players "Too many players")
+
+(defn ini-hand
```
kenan-rhoton: WOAH! You made a random rules change, can you please make this a vector of `{:power 10}`. Although it `should` be enough to make the change, I would rather have it isolated in another issue. Also, is there a test for this?

In any case, please make this be {:power 10} for now.

backend/test/card_game/api_play_test.clj
```diff
@@ -0,0 +1,55 @@
+(ns card-game.api-play-test
+  (:require [expectations.clojure.test :refer :all]
+            [card-game.api.base :as api]
+            [configs :as configs]))
+
+(defexpect playing-cards
+  (expect
```
kenan-rhoton: `; Cards in hand have power`

It took me a moment to read what the expectation was looking for and I'm pretty sure I wrote the test 🙃

backend/test/card_game/api_play_test.clj
```diff
@@ -0,0 +1,55 @@
...
+    #(= (count (:hand %)) (dec (count (configs/ini-hand))))
+    (let [game (api/create-game)
+          opponent (api/add-player (:game-id game))]
+      (do
+        (api/play-card-as-player (:game-id game) (:player-id opponent) 0 0)
+        (api/play-card-as-player (:game-id game) (:player-id game) 0 0)
+        (api/get-game (:game-id game) (:player-id game)))))
+
+  ; card played is owned by self
```
kenan-rhoton: This test and the next one can be merged into one "Card is owned by the player that played it"

Two approaches to do it:

1. Extract the `(do` bodies into a `(let` with both the expects *inside* the `(let` (there is no reason `(expect` needs to be a direct child of `(defexpect` I believe.
2. Make one lone expect with both acceptance functions inside an `(and` (but we would then not now which of the two cases failed upon a failure)

backend/test/card_game/api_winner_test.clj
```diff
@@ -0,0 +1,44 @@
+(ns card-game.api-winner-test
+  (:require [expectations.clojure.test :refer :all]
+            [card-game.api.base :as api]
+            [configs :as configs]))
+
+(defexpect win-conditions
```
kenan-rhoton: All these three expectations could be pretty much two-liners if we simply make ourselves a helper function above this
```clojure
(defn play-all-cards
  [player1 player2 strategy1 strategy2]
     ...
```

backend/test/card_game/core_base_test.clj
```diff
@@ -0,0 +1,51 @@
...
+
+  ; Hand's not empty
+  (expect
+    false
+    (-> (core/new-game)
+        (get-in [:players 0 :hand])
+        (empty?))))
+
+(defexpect points-on-rows
+  ; We can get the current amount of points on different rows
```
kenan-rhoton: These functions really look like they could be simplified... I don't know how right now, though.

backend/test/card_game/effects_test.clj
```diff
@@ -1,33 +1,48 @@
...
   (:require [expectations.clojure.test :refer :all]
-            [card-game.core :as core]))
+            [card-game.core :as core]
+            [configs :as configs]))
 
 (defexpect card-altering
   ; Altering a card in hand works
   (expect
-    [15 10 10 10 10 10 10 10 10 10 10 15]
+    (concat [9001] (subvec (vec (map :power (configs/ini-hand))) 2))
```
kenan-rhoton: I don't understand why we play a card at all in this test if we only need to check that altering a card works

backend/test/card_game/effects_test.clj
```diff
@@ -1,33 +1,48 @@
...
-             (core/alter-card [:players 0 :hand 0] {:power 15})
-             (core/alter-card [:players 0 :hand 11] {:power 15})
+             (core/alter-card [:players 1 :hand 0] {:power 9001})
+             (core/play-card 0 0 0)
+             (core/play-card 1 1 1)
+             :players
+             (nth 1)
+             :hand)))
+
+  (expect
```
kenan-rhoton: I think a few `(let` would make this test much more readable. (for example setting `last-card (dec (count (configs/ini-hand)))`

backend/test/card_game/simultaneous_test.clj
```diff
@@ -0,0 +1,60 @@
+(ns card-game.simultaneous-test
+  (:require [expectations.clojure.test :refer :all]
+            [card-game.core :as core]
+            [card-game.victory-conditions :as victory-conditions]
+            [configs :as configs]))
+
+(defexpect ot-of-turn
```
kenan-rhoton: o**u**t-of-turn 🙃

backend/test/card_game/test_helper.clj
```diff
@@ -0,0 +1,20 @@
+(ns card-game.test-helper
+  (:require [card-game.core :as core]
+            [configs :as configs]))
+
+(defn end-game
+  ([game-state] (end-game game-state 0))
```
kenan-rhoton: if we base the exit condition on the current number of cards in hand extracted from the game-state, we don't need the previous-plays parameter.

backend/test/card_game/winner_test.clj
```diff
@@ -0,0 +1,48 @@
...
+
+  ; Winner returns the winning player on a finished game, or 2 on a tie
+  (expect
+    0
+    (-> (core/new-game)
+        (core/play-card 0 0 3)
+        (core/play-card 1 0 2)
+        (core/play-card 0 0 0)
+        (core/play-card 1 0 2)
+        (helper/end-game 2)
```
kenan-rhoton: Ok, this is beautiful (but I still believe we shouldn't need the `2`)

backend/src/card_game/api/base.clj
Masclins: Right now `add-player` returns `{:error configs/too-players}` when too many players.
Making the check on `helper/create-players` (without many other changes) would imply calling `(get-game` with an `{:error configs/too-players}` game-state. I think the result of that might be a cause of bugs.

backend/src/card_game/api/helper.clj
Masclins: As explained on the previous comment, I disagree.

backend/src/card_game/core.clj
Masclins: We were not able to do so without making several other changes.
Since, right now, hands are defined on `configs/ini-hand` I see no reason to have that.
At some point we will implement different hands and then will be the time to work this out. Having a structure like the one we had doesn't help either for that.

backend/src/card_game/api/base.clj
kenan-rhoton: 🤔 maybe... I don't know, `add-player` is one function I'm not very happy about, I just feel like it does too many things:

1. Fetches a game
2. Checks how many players the game has and returns an error if there are too many
3. If the game doesn't exist, it creates it.
4. It adds a player to the game.
5. It fetches the game as a player and returns it. <-- This I will make an issue to change accross the board so ignore it

I feel like 2 and 3 shouldn't be the responsibility of a function called `add-player`

backend/src/configs.clj
Masclins: It is as intended since it's the message that says the player she needs to play.

backend/src/configs.clj
Masclins: same here too

backend/src/configs.clj
Masclins: The issue was to include all those changes on configs.clj.
Being able to change the file while passing the test is what I understood as the test itself. I'll change it, but I needed to check I was able to change it without any problem.

backend/src/card_game/api/base.clj
Masclins: I feel like `add-players` should check if you can actually add that player, so I don't see a problem with 2.
Agree with 3.

kenan-rhoton: Ok, I can agree to that.

backend/test/card_game/api_play_test.clj
Masclins: I understand that's because you don't like to testes that are *the same game* with different expects. Still, I don't think it will then be more readable doing that.
Anyway, I'll add more stuff to the `let` of all those tests.

kenan-rhoton: This one is not about readability, it's about duplication. If we change, for example, how games are generated, we would need to duplicate the change here unnecessarily, since we really use the same game.

backend/test/card_game/api_winner_test.clj
Masclins: I thought about something like that, but wanted to keep the `helper/end-game` as simple as possible. If you think it's worth I'll work something out.

backend/test/card_game/effects_test.clj
Masclins: any suggestion will be welcomed

backend/test/card_game/test_helper.clj
Masclins: I don't want to use the game-state because we are (sometimes) testing that the game-state correctly tracks the number of cards in hand.

backend/test/card_game/winner_test.clj
Masclins: As before, I want to avoid using the game-state for test that are checking the game-state.

backend/test/card_game/api_winner_test.clj
kenan-rhoton: `helper/end-game` could be redefined to use this function internally, too, so it wouldn't really add much complication

backend/test/card_game/test_helper.clj
kenan-rhoton: But that shouldn't be what this function does, right? I think the purpose of the function is to drive the game to an end no matter what. And I don't recall seeing any test where we use this function with an incorrect number of previous plays (we might, I just don't recall it)

Masclins: Will do that. Maybe that also helps adding strategies

backend/test/card_game/api_winner_test.clj
kenan-rhoton: This is what I meant:

```clojure
(defn play-a-game-helper
  [strategy1 strategy2]
  (loop [game (api/create-game)
         opponent (api/add-player (:game-id game))
         iteration (count (configs/ini-hand))]
      (if (= 0 iteration)
        (api/get-game (:game-id game) (:player-id game))
        (recur
          (api/play-card-as-player (:game-id game) (:player-id game) 0 (strategy1 iteration))
          (api/play-card-as-player (:game-id game) (:player-id opponent) 0 (strategy2 iteration))
          (dec iteration)))))
    

(defexpect win-conditions
  ; Both stack all cards on one row -> tie
  (expect
    #(= :tie (:winner %))
    (play-a-game-helper
      (fn [i] 0)
      (fn [i] 0)))

  ; Opponent stacks all cards on one row and I spread -> I win
  (expect
    #(= :me (:winner %))
    (play-a-game-helper
      (fn [i] (mod i 4))
      (fn [i] 0)))

  ; I stack all cards on one row and opponent spreads -> Opponent wins
  (expect
    #(= :opponent (:winner %))
    (play-a-game-helper
      (fn [i] 0)
      (fn [i] (mod i 4)))))
```

backend/src/card_game/api/helper.clj
kenan-rhoton: skipped this one 🙃

PR #38, opened 2018-03-01
Frontend: Improved readability of Game component

### LINKED ISSUE #29: `dropOnRow` is really confusing
`dropOnRow` on **frontend/src/components/Game.vue** is really confusing and requires an IQ of over 9000 to fully understand.

kenan-rhoton:
I'm hoping this is enough to resolve #29 🙃

### DIFF
--- a/frontend/src/components/Game.vue
+++ b/frontend/src/components/Game.vue
@@ -58,21 +58,14 @@ export default {
   },
   methods: {
     gameReq: function() {
-      return (
-        "http://backend:3000/games/" +
-        this.$route.params.gameID +
-        "/player/" +
-        this.$route.params.playerID
-      );
+      const gameID = this.$route.params.gameID;
+      const playerID = this.$route.params.playerID;
+      return `http://backend:3000/games/${gameID}/player/${playerID}`;
     },
     updateGame: async function() {
-      try {
-        const req = this.gameReq();
-        const response = await axios.get(req);
-        this.response = response.data;
-      } catch (err) {
-        throw err;
-      }
+      const req = this.gameReq();
+      const response = await axios.get(req);
+      this.response = response.data;
     },
     dragCardFromHand: function(ev) {
       ev.dataTransfer.setData("handIndex", ev.target.getAttribute("index"));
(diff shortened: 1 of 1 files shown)


frontend/src/components/Game.vue
```diff
@@ -58,21 +58,14 @@ export default {
...
     gameReq: function() {
-      return (
-        "http://backend:3000/games/" +
-        this.$route.params.gameID +
-        "/player/" +
-        this.$route.params.playerID
-      );
+      const gameID = this.$route.params.gameID;
+      const playerID = this.$route.params.playerID;
+      return `http://backend:3000/games/${gameID}/player/${playerID}`;
```
Masclins: This is way clearer now.

frontend/src/components/Game.vue
```diff
@@ -58,21 +58,14 @@ export default {
...
     },
     updateGame: async function() {
-      try {
-        const req = this.gameReq();
```
Masclins: Why removing the try/catch? Or why was it here the first time?

kenan-rhoton: it was there to catch the exception possibly thrown by a connection error in axios.get... but then I changed the catch to just rethrow the error... and then I realized, if I'm just rethrowing the error, why is the `try` even there?

frontend/src/components/Game.vue
```diff
@@ -84,27 +77,24 @@ export default {
       ev.preventDefault();
       const req = this.gameReq();
       const data = ev.dataTransfer.getData("handIndex");
-      const sentData = { index: data, row: ev.target.getAttribute("rownum") };
-      axios.post(req, sentData);
+      const playData = {
```
Masclins: More readable.
I like using more comprehensive names.

frontend/src/components/Game.vue
```diff
@@ -84,27 +77,24 @@ export default {
...
-      return (
-        loc.protocol +
-        "//" +
-        loc.hostname +
-        ":" +
-        loc.port +
-        "/#/join-game/" +
-        this.$route.params.gameID
-      );
+      const l = window.location;
```
Masclins: How is `l` better than `loc`?

kenan-rhoton: it's literally used only two lines down and I only define the constant to make the code shorter and fit in one line so it's easy to read the format

Issue #43, opened 2018-03-02
Centralize our configuration


Masclins:
To follow DYR, most (if not all) error and status messages are on `backend/src/configs.clj` (maybe it should be nested somewhere else?)

There are still some of those hardcoded on `frontend/test/unit/Game.spec.js` (lines 16, 26 and 29).
(I couldn't find any other case where that happens)

Those should make reference to what's on `backend/src/configs.clj`.

Masclins:
As per #44 configs should be on different files, in `backend/src/configs/`.

Masclins:
How is this an "enchancement"? Isn't it a problem that a file doesn't do what it's supposed to do? (In this case: `configs/messages.clj` including all the messages to DRY and letting us change those)

kenan-rhoton:
Bug: Something isn't doing what we programmed it to do.
Enhancement: We want our program to reflect new design decisions.

kenan-rhoton:
I've run into a little issue that we should all be aware of, and which I hadn't thought about:

1. Config shared between `backend` and `e2e` is solved
2. However, having `frontend` have access to `config` means that it will be 100% public (as the file will get loaded by the browser).

This shouldn't be at all a problem, but we should all be aware that anything in the config file will be considered 100% public.

I'll put a comment to remind us of it at the top.

Masclins:
It should have access to certain files on `config`, not all of them.
Right now, only `messages.clj`.

PR #49, opened 2018-03-05
Vueless proposal

### LINKED ISSUE #45: Kill Vue with fire
It's an unnecessary dependency in our project (caused by me) that provokes:

1. Weird frontend structure
2. Less intuitive and readable code in the frontend
3. Less flexibility to do things like #44 
4. No great advantage (it does provide some) to making our game better

We should instead apply the same philosophy we do to backend: no new dependencies until it's absolutely unavoidable, and attempt to keep everything as clean, organized and simple as possible.

If we do this, we'll have to stop and close all frontend issues to undertake this, and it would be a team effort (I can split the tasks among different issues).

It would have the added benefit that if we all actually work on the frontend base, we'll better understand it, especially if we try to avoid "magic libraries" for as long as possible.

kenan-rhoton:
This is my initial proposal to close #45 

Still requires linting, dividing up game.js and testing, probably with chai.js

Vote as you will


Masclins:
I see this as clearear than with Vue. And easier to know how to find what something does (since it's always Javascript).

The files called `index.html` must be called that way? Either explaining why they are called *index* or changing them to a more comprehensive name would make it easier (at least for me).

kenan-rhoton:
There are two options if we rely on simple `http.server` protocols:

1. use a clearer name than index.html and have something like `/lobby/join.html?gameID=5` which looks very unprofessional nowadays due to the file extension being part of the URL.
2. rely on the default for a folder `/lobby/join/` redirecting to `/lobby/join/index.html` silently, which we currently do.

Other options would be:

1. Not using plain `http.server` and instead programming a small implementation of it (probably under 80 lines of Python/Go/JavaScript or under 200 if we want to include templating)
2. Using an existing server framework (completely unnecessary bloat in my opinion, but I may be persuaded)
3. Have backend serve frontend (which completely breaks the entire backend/frontend separation of concerns, but might be slightly less work than point 2)

Masclins:
I'm ok using `index.html` as long as I understand what it is and why we use that name (and I do now)

kenan-rhoton:
Things still missing:

- Readability pass
- Tests (mainly because this is just a proposal, but we *cannot* merge this without them)

For tests I propose either using plain `assert` or simply importing `chai.js` which is a fairly easy library to use for tests.

kenan-rhoton:
Ok, we will now focus or testing on the e2e side, frontend is basically non-unit-testable (and probably should be kept that way by minimizing any logic).

In the future, if we need to test logic in frontend, we will employ Offensive Programming with inline `assert`s.

frontend/lobby/waiting/waiting.js
```diff
@@ -0,0 +1,34 @@
...
+
+function checkForOpponent() {
+    var req = new XMLHttpRequest();
+    req.open("GET", `http://backend:3000/games/${gameID}/player/${playerID}`);
+    req.responseType = "json";
+    
+    req.onload = function () {
+        const status = req.response["status"];
+        
+        if (status !== "Waiting for an opponent") {
```
Masclins: Couldn't we solve #43 on this merge?

frontend/game/board/drop.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+
+define(function (require) {
+    var helper = require('../helper.js');
+
+    return {
+        allow(event) {
+            event.preventDefault();
+        },
+        onRow(event) {
```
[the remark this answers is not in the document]
Masclins: Seeing `event.target.getAttribute("rownum")` I think I answered the question myself...

frontend/game/hand.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+define(function(require) {
+    var helper = require('./helper.js');
+    return {
+        setHand(handState) {
```
Masclins: Is this called pretty often or only once?
Because it feels like a slow browser could be showing the hand to disappear and reappear... (maybe I'm just saying silly things)

frontend/game/helper.js
```diff
@@ -0,0 +1,17 @@
+"use strict";
+
+let params = new URLSearchParams(document.location.search.substring(1));
+
+define(function() {
+    return {
+        clear(elem) {
```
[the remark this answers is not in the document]
kenan-rhoton: because I'm an idiot

frontend/lobby/waiting/waiting.js
kenan-rhoton: That is a whole different lot of work that will imply both backend and frontend changes and probably some sort of `config.yml` file at the root.

frontend/game/hand.js
kenan-rhoton: Check out https://github.com/kenan-rhoton/card-game/pull/49/files/7ecddddebdc458687aac0a8f7ccc29a4bcf1f806#diff-3075d78b5d6ef0ac1276cb712385903fR18

Masclins: So it only reloads it when the number of cards changed?
Will need change if we create any new rule that makes you "draw" new cards, but right new seems like a nice solution.

Issue #44, opened 2018-03-03
File reorganization


kenan-rhoton:
Ok, I think we need to reorganize and clarify the project structure, this is what we currently have:

- **backend**
    - **doc**: why does this even exist?
    - **src**: Where the backend production  code lives
        - **card_game**: Where the backend production code lives, except configs.clj for some reason.
            - **api**: Were we define how frontend or any other client can communicate with the system
    - **test**: Where the backend test code lives.
        - **card_game**: Where the backend test code lives.
- **e2e**: Where the end-to-end tests live.
- **frontend**
    - **public**: Where the... um... index.html lives?
        - **img**: Images?
            - **icons**: icons?
    - **src**: Where the frontend javascript production code lives.
        - **components**: Where the javascript... components... live?
        - **views**: I give up, what's the difference, again?
    - **test**: Where the frontend tests live.
        - **unit**: Where the frontend unit tests live

I feel like we could do much better and have a clearer organization with something like:

- **backend**
    - **src**: Where the backend production code lives
        - **rules**: Where are the rules logic lives (currently "core")
        - **api**: Where we define how frontend or any other client can communicate with the system
        - **persistence**: Where we define how we store persistent data
        - **configs**: Where we load our configuration
    - **test**: Where the backend test code lives.
        - **api**: Api tests
        - **rules**: Rules tests
- **e2e**: Where the end-to-end tests live.
- **frontend**
    - **static**: Where static stuff (images, stylesheets, whatever) lives.
    - **src**: Where the frontend production code lives
        - **lobby**: Where all the Game Creation/Game Joining logic lives
        - **game**: Where the actual gameplay occurs
    - **test**: Where the frontend test code lives
        - **lobby**: Where we do all the lobby tests
        - **game**: Where we do all the game tests

Any thoughts or ideas? This would be a pretty big change, and since it is about understanding the system and having everything in a clear place, I would like *everyone* to say something about it 🙃

Masclins:
I like that.
Still we should wait until we close the #39 pull-request that also changed organitzation, shouldn't we?

Configs would be a directory with multiple config files on it (probably clearer)?

kenan-rhoton:
This would be post #39, yes 🙃

Masclins:
I'll begin working with the backend of this, waiting for #39 before touching any affected file by that pull-request.

When that's done, if nobody begun with front-end, I'll do it too.

kenan-rhoton:
[...]

PR #50, opened 2018-03-06
File reorganization of backend

Masclins:
This solves only half #44: The backend part.

Frontend part will wait until we decide what to do with #45.


Masclins:
@ruchnar this can be a good one to review to understand how `(ns` and `(require` works.

PR #51, opened 2018-03-06
Function to get rows won for each player

### LINKED ISSUE #47: New function needed: get-total-points
Right now, under /backend/src/card_game/victory_conditions.clj we have these two functions: 

get-points -> Tells how many points each player has on each row
most-points -> Tells which player has the most points

We need to add another one as a step to solve the issue #18 . 

The function should be named "**get-total-points**", it should return the number of rows won for each player (eg the number of points for each player), and it should be placed between **get-points** and **most-points**.

Also, we should change how **most-points** is coded in order to re-use the content of the new function and simplicity.

Masclins:
This closes #47. The function is called `get-won-rows`.

Also, used the new function to simplify `most-points`. 

After this update, `get-points` is only used on test. We should reconsider what to do with it.


kenan-rhoton:
never be afraid to remove useless functions like `get-points`. Since git saves all history, we can always revert the change if it's a mistake 😉

Masclins:
I don't dislike having `get-points` because it lets us do many tests that would be complex otherwise.

Masclins:
After checking which tests use `get-points` I think it's better to remove it and adapt some of those tests.

backend/src/rules/victory_conditions.clj
```diff
@@ -18,40 +18,33 @@
...
       row-data)))
 
-(defn get-points
-  "Tells us how many points each player has on each row"
-  [game-state]
-  (loop [row 0
-         points []]
-    (if (> row 4)
-      points
+(defn get-won-rows
```
kenan-rhoton: two things: 

1. I would define `(let [opponent (mod (inc player) 2)]` at the beginning of the function.
2. I think this would look better as a `reduce` than a `loop`

backend/test/rules/winner_test.clj
```diff
@@ -29,6 +29,36 @@
       (fn [i] 0)
       (fn [i] 0))))
 
+(defexpect row-winner
```
kenan-rhoton: This file is getting long 🙃

kenan-rhoton: Yes, we still need to watch it, and it might be useful to split on certain conditions (like draws, player1-wins, player2-wins or whatever)

backend/src/rules/victory_conditions.clj
Masclins: some clue regarding how to do that? :)

kenan-rhoton: This quick solution requires redoing `points-in-row`:

```clojure
(defn points-in-row
  "Tells us how many points a certain row contains for a certain player"
  [game-state row]
  (reduce
    #(if (= (:owner %2) player)
       (+ %1 (:power %2))
       %1)
    0
    row))

(defn winning-player?
  [row player]
  (let [opponent (mod (inc player))]
    (> (points-in-row row player)
       (points-in-row row opponent))))

(defn get-won-rows
 "Tells us how many rows a player is winning"
 [game-state player]
 (reduce
   #(if (winning-player? %2 player) (inc %1) %1)
   0
   (get-in game-state [:rows])))
```

By the way, I didn't find a way to quickly chech if points-in-row is used elsewhere. There's one thing we should **really** start doing, and it's marking private functions as private by declaring them with `(defn-` which dissallows using them outside their namespace. That way we know which functions are only extracted logic and don't need to be tested.

kenan-rhoton: Also, if preferred, `(defn ^:private points-in-row` is longer, but clearer, and for some reason, the only way to do private `(def`s ( there is no `(def-`)

Issue #35, opened 2018-02-28
Playing cards by clicking


Masclins:
Right now cards can only be played by dragging them onto the boardgame.
They should be playable *also* by clicking them and then clicking on the row.
That would imply marking them while they are "selected" (Probably should be worked with #16, for using an outline in both cases, but with different color).

ruchnar:
I don't know how to implement this with clojure, but there should be an :on-click function... somewhere...

Masclins:
This is *forntend-only* so nothing needs to be done in clojure.

Issue #16, opened 2018-02-25
Clarify next-play


kenan-rhoton:
Add a visualization for the next-play (probably by putting a different background color on the card or outline or something).

Right-now there is no indication that you even made the play correctly.

Masclins:
Writing the status somewhere would be useful too ("Play a card" or "Wait for your opponent's play")

Masclins:
I see two ways to do that:

- Adding something like `:status "next-play"` on cards to be played on backend
- Checking `:next-play`. If doing the second, shouldn't frontend only recieve the index of the card?

Either way it doesn't seem like a *frontend-only* issue.

kenan-rhoton:
I feel like the second option is better: backend already gives enough information for what we need.

And agreed that frontend should only receive the index and only if the `:next-play` is his.

kenan-rhoton:
I also think this and #35 should be done together.

PR #62, opened 2018-03-08
Cards can be played by clicking, solving #35

Masclins:
This lets playing by clicking card and row afterwards.

It also highlights the selected card (that's going to be played).

Since this is really tied to #16, I'd like a review before moving into it, so I go there with the code as clear as possible.


frontend/game/board/board.js
```diff
@@ -32,5 +32,22 @@ define(function (require) {
         },
         allowDrop: drop.allow,
         dropOnRow: drop.onRow,
+        clickRow(event) {
+            if (helper.clickedCard != -1) {
```
kenan-rhoton: Please extract the playCard logic so this and drop.onRow can share a function

frontend/game/hand.js
```diff
@@ -2,6 +2,23 @@
...
+    function clearBackground() {
+        document.querySelectorAll(".card").forEach(function(element) {
+            element.style.background = "white"
+        });
+    }
+
+    function clickCard(card) {
+        clearBackground();
+        if (helper.clickedCard == card) {
+            helper.clickedCard = -1;
```
kenan-rhoton: please use `undefined` instead of `-1`

frontend/game/helper.js
```diff
@@ -2,17 +2,33 @@
...
         },
-        clickedCard: -1,
-        gameID: params.get("gameID"),
-        playerID: params.get("playerID"),
+        clickedCard: undefined,
+        gameID: gameID,
+        playerID: playerID,
         baseCard: document.getElementById("card-template")
                 .content.querySelector(".card"),
+        playCard(rownum, cardindex) {
```
kenan-rhoton: I think this should be in board/play.js rather than helper 🤔

kenan-rhoton:
Ok, I see 2 options now: either we move `play.js` to be inside `board` or we include the `hand` functions that have to do with playing a card in it.

There's also a third option of having a new `game/play` directory with a file for the `board` functions and another for the `hand` functions 🙃

Masclins:
Tests pass in my computer but seems not to pass for Travis...

Masclins:
About what to do with `play.js`:

1. *`play.js` inside `board/`*: This feels weird to me. Playing a card is not something specific to the board.
2. *Move some functions in `hand.js` to `play.js`*: I assume you talk about clicking and dragging a card. This seems somewhat ok, though I'm not sure if it really makes anything clearer. And then probably should move the equivalents from `board.js`. And by doing so `play.js` would end up being a mess.
3. *Having `game/play/` with `play.js`, `hand.js` and `board.js`*: If we have to choose one, this has the most sense to me. My only question is if we really want to divide files that much.

So, in my oppinion, we either leave it as it is or do the third option.

kenan-rhoton:
Go with option three then, I will fix travis in parallel

frontend/game/index.html
```diff
@@ -11,7 +11,7 @@
         <div id="app" class="text-center container">
             <h1 class="display-1">
                 <a href="/">
-                    Troll Card Game
+                    TCG
```
Masclins: I really hope you are ok with this.
It's too NSFW to have **Troll Card Game** on screen when testing something.

frontend/game/hand.js
```diff
@@ -28,7 +13,7 @@ define(function(require) {
                 var newCard = helper.baseCard.cloneNode(true);
                 newCard.innerHTML = cardInHand["power"];
                 newCard.setAttribute("index", index);
-                newCard.addEventListener('click', function(){clickCard(newCard)});
+                newCard.addEventListener('click', function(){play.clickCard(newCard)});
```
Masclins: I have the feeling this could be prettier, but don't know how.

frontend/game/play/play.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
-define(function() {
-    var helper = require('./helper.js');
+define(function(require) {
```
Masclins: Maybe this is why tests failed.

frontend/game/hand.js
kenan-rhoton:
```javascript
newCard.addEventListener('click', () => play.clickCard(newCard) );
```
maybe?

frontend/game/play/play.js
kenan-rhoton: nah, that's a separate issue I have to fix

frontend/game/index.html
kenan-rhoton: good idea

frontend/game/hand.js
Masclins: I saw this alternative on SE.
But it seems equally ugly + harder to understand.

PR #65, opened 2018-03-12
Clarifies next-play

Masclins:
Solves #16 by:

1. Marking card in yellow when playing it.
2. Showing status on top.
3. Not letting click nor drop (for playing) when status is not *play*.

### DIFF
--- a/e2e/test.js
+++ b/e2e/test.js
@@ -76,15 +76,19 @@ test('Sample Game', async testCase => {
 
     while (await gamePage.cardsInHand.exists) {
         await gamePage.checkState(testCase, maxHandNum, cardsInPlay)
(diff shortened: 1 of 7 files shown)


Masclins:
There are some issues that arose while solving this issue. Maybe some should be solved on this pull-request, though maybe they are worth a new issue.

- `helper.js` is begining to stop being a simple *helper*. Maybe the newly added functions should be on a different file?
- Having `board/board.js` and `play/board.js` or `hand.js` and `play/hand.js` might be confusing.
- This stops showing the next-play upon refreshing. But including everything here on `setState` maybe should imply changing a bit how it works and when it's called.

Issue #63, opened 2018-03-08
Give options when creating game


Masclins:
Right now `rules/create_game.clj` always use `hand/ini-hand` for the initial hand.

We are planning row limit (#42) and power-changing cards (#41), that we will need to put into the game.

But trying to make all the tests handle all the possible cards and rows configurations will be *impossible*.

For that I suggest being able to have some options when creating new-games (maybe having different configs to test different things?). That will allow us to design tests targeting what we wanna test, and not spending a lot of effort to solve problems not related to the test.

Probably this Issue should be addressed before #42 and #41.

PR #66, opened 2018-03-12
Can give conditions upon game-creation (Solves #63)

Masclins:
Some changes in `config/` names.

Now `new-game` can be called with a map of attributes (right now only `:hand` is used) to describe the initial conditions. Non described use the `default` options.

This solves #63, which will help doing the tests for the rule-changing issues.

### DIFF
--- a/backend/src/configs/hands.clj
+++ b/backend/src/configs/hands.clj
@@ -1,8 +1,7 @@
-(ns configs.hand)
+(ns configs.hands)
 
-(defn ini-hand
-  "Generates de initial hand (can't be empty)"
-  []
+(def default-hand
+  "Default hand (can't be empty)"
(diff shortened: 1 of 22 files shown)


backend/src/configs/hands.clj
```diff
@@ -1,8 +1,7 @@
-(ns configs.hand)
+(ns configs.hands)
 
-(defn ini-hand
-  "Generates de initial hand (can't be empty)"
-  []
+(def default-hand
```
kenan-rhoton: Why change from `defn` to `def`?

Masclins: It wasn't a function.
Since I had to change them all anyway, I thought I could change that too.

PR #67, opened 2018-03-13
Make everything deployable

### LINKED ISSUE #54: Make the system deployable and playable online
We need to be able to play among ourselves!

This will mostly be a DevOps style task, and I'm not sure yet if we'll put it on v0.2.0 or v0.3.0, but I want it to be visible!

kenan-rhoton:
Closes #54 
A bunch of DevOps changes to:

1. Configs
2. Travis

To make everything easy to deploy and letting us play and test this online.


configs/config.yml
```diff
@@ -5,3 +5,6 @@ messages:
     play: "Playing"
     wait: "Waiting for opponent's play"
     too-many-players: "Too many players"
+servers:
+    frontend: "frontend:8080"
+    backend: "backend:3000"
```
Masclins: Where will backend be?

configs/config.yml
```diff
@@ -5,3 +5,6 @@ messages:
     play: "Playing"
     wait: "Waiting for opponent's play"
     too-many-players: "Too many players"
+servers:
+    frontend: "frontend:8080"
```
Masclins: This, I assume, will be a route to some downloaded part from the player?

frontend/lobby/join/join.js
```diff
@@ -3,10 +3,11 @@
 let params = new URLSearchParams(document.location.search.substring(1));
 const gameID = params.get("gameID");
 
-document.addEventListener('DOMContentLoaded', function() {
+onConfigLoad(function() {
```
Masclins: I did not understand this part before either. What does it do?

.travis.yml
```diff
@@ -1,12 +1,29 @@
 sudo: required
```
Masclins: Seems you'll have to explain a lot before being approved.

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
         build: frontend
         restart: always
         ports:
-            - "8080:8080"
+            - "8880:8080"
```
Masclins: Now `localhost:8501` will work too?

kenan-rhoton: if you change it here AND in config/config.yml it should work (I still haven't found a way to unify these two places)

frontend/lobby/join/join.js
kenan-rhoton: This executes the function passed to it when config is done loading (the function is defined in frontend/config/config.js), to avoid referencing the configuration before it's loaded.

Issue #58, opened 2018-03-07
Scores on each row


Masclins:
Now that we show the number of rows won, it would be great to show how many points each player has on each row.

It would be great (not sure if it should be another Issue) to also visually show who's winning for each row, maybe by highlighting his number (and the same for the number of rows won).

kenan-rhoton:
This will not go on any current milestones, but if anyone wants to do it, feel free.

Masclins:
After playing (against myself) I feel like the scores part is necessary for having useful feedback from playtesting.

Otherwise it feels more like "who gets tired before of adding numbers"

(maybe we could put the highlighting part into another Issue and work only with the scores for v0.2.0)

kenan-rhoton:
Ok, adding to milestone

Masclins:
Why Issues are now longer auto-closed?

kenan-rhoton:
You have to write "solves #58" in the pull request *description* not the title

PR #70, opened 2018-03-14
Scores on rows, solving #58

Masclins:
Right now it is extremely ugly and can't manage to make it centered.

Any suggestion on how to change that will be welcomed.


kenan-rhoton:
I think nothing should be called helper, now that I think about it, that's just a "disaster area" to put anything you want in, instead of a description of a functionality as it should be...

backend/src/api/helper.clj
```diff
@@ -29,6 +29,21 @@
...
+  [game-state player-id]
+  (let [player (player-num game-state player-id)
+        opponent (mod (inc player) 2)]
+    (loop [rows-power []
+           rows (:rows game-state)]
+      (if (empty? rows)
+        rows-power
+        (recur
+          (conj rows-power
+                [(victory-conditions/points-in-row (first rows) player)
```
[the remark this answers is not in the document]
kenan-rhoton: also, we should rename this file to `player_view` and move any functions below to somewhere else (and what does `define-status` even do?)

backend/src/rules/victory_conditions.clj
```diff
@@ -7,7 +7,7 @@
     (= 0 (count (get-in game-state [:players 0 :hand])))
     (= 0 (count (get-in game-state [:players 1 :hand])))))
 
-(defn ^:private points-in-row
+(defn points-in-row
```
kenan-rhoton: nice function reuse! 😃

backend/test/api/scores_test.clj
```diff
@@ -0,0 +1,55 @@
+(ns api.scores-test
+  (:require [expectations.clojure.test :refer :all]
+            [api.base :as api]
+            [test-helper :as helper]
```
kenan-rhoton: once again, we need a more descriptive name than `helper` (could be another issue)

kenan-rhoton: I suggest "auto-play" for that file, instead of `test-helper`

backend/test/api/scores_test.clj
```diff
@@ -0,0 +1,55 @@
...
+  ; Game begins with scores at 0
+  (expect
+    [[0 0] [0 0] [0 0] [0 0] [0 0]]
+    (let [game-state (api/create-game)
+          game-id (:game-id game-state)
+          player-id (:player-id game-state)
+          opponent-id (:player-id (api/add-player game-id))]
+      (:rows-power (api/get-game game-id player-id))))
+  (expect
+    [[(helper/default-hand-power 0) (helper/default-hand-power 1)]
```
kenan-rhoton: both `helper` and `default-hand-power` aren't very descriptive, I think, since I believe this generates a card of power 0 (correct me if I'm wrong)

backend/test/api/scores_test.clj
```diff
@@ -0,0 +1,55 @@
...
+    (let [game-state (api/create-game)
+          game-id (:game-id game-state)
+          player-id (:player-id game-state)
+          opponent-id (:player-id (api/add-player game-id))]
+      (do
+        (api/play-card-as-player game-id player-id 0 0)
+        (api/play-card-as-player game-id opponent-id 1 0))
+      (:rows-power (api/get-game game-id player-id))))
+  (expect
+    [[0 (+ (helper/default-hand-power 0) (helper/default-hand-power 1))]
```
kenan-rhoton: ok, now you're adding them together, I think the name is definitely not descriptive enough

backend/test/test_helper.clj
```diff
@@ -1,6 +1,37 @@
 (ns test-helper
```
kenan-rhoton: definitely let's rename this `auto-play`, please

backend/test/test_helper.clj
```diff
@@ -1,6 +1,37 @@
...
+  (:require [configs.hands :as hands]
+            [api.base :as api]
+            [rules.create-game :as create-game]
+            [rules.play-card :as play-card]))
 
 (defn default-hand-power
   [x]
   (get-in hands/default-hand [x :power]))
+
+(defn play-as-api
```
kenan-rhoton: I feel these two functions can be abstracted together, maybe by using a third function to build on.

backend/test/api/scores_test.clj
Masclins: You're wrong.
This returns the power of the nth card in the `default-hand`

backend/test/api/scores_test.clj
Masclins: Since you don't understood what they do it's probably not descriptive enough. Even though I can't think of a better name (that's not way longer).

`nth-default-card-power`?

backend/test/api/scores_test.clj
kenan-rhoton: ok, absolutely definitely change the name to something understandable, please, please, like `hand/power-of-nth` or something

Masclins: `hand/power-of-nth index hand`, so it can be also reused for different hands (we will have them pretty soon, when introducing power-changing abilities).

backend/src/api/helper.clj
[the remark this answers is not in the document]
Masclins: ...oddly enough it's added on `get-game` rather than on `get-game-as-player`...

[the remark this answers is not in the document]
kenan-rhoton: ... ok, that's not right at all

backend/test/hand.clj
```diff
@@ -0,0 +1,6 @@
+(ns hand
```
kenan-rhoton: test.hand

backend/test/hand.clj
```diff
@@ -0,0 +1,6 @@
...
+  (:require [configs.hands :as hands]))
+
+(defn default-hand-power
```
kenan-rhoton: also, forgot to rename to power-of-nth

backend/src/api/helper.clj
```diff
@@ -49,6 +64,7 @@
    :player-id player-id
    :hand (get-hand game-state player-id)
    :rows (get-rows game-state player-id)
+   :rows-power (get-rows-power game-state player-id)
    :scores (get-scores game-state player-id)
    :winner (get-winner game-state player-id)})
```
kenan-rhoton: I still think any function below this either should be calculated inside `get-game-as-player` or should be in a different file, with this one being named `api.player-view` or something like that

frontend/game/board/builder.js
```diff
@@ -17,6 +21,16 @@ define(function (require) {
 
                 var oppRow = baseRow.cloneNode(true);
                 oppRows.appendChild(oppRow);
+
+                var newScore = baseScore.cloneNode(true);
+                newScore.setAttribute("rownum", i);
+                newScore.innerText="1";
```
kenan-rhoton: why the hard-coded "1" and "2" assigned to `innerText` and `innterHTML`? (and why use a different attribute, anyway?)

Masclins: Whoops...
I used that for testing, and it didn't matter after coding the rest, so I forgot them there.

backend/src/api/helper.clj
Masclins: I was not sure about the `api.player-view-functions` but at least it's a bit clearer now.

backend/src/api/base.clj
```diff
@@ -1,24 +1,22 @@
 (ns api.base
   (:require [rules.create-game :as create-game]
             [rules.play-card :as play-card]
-            [rules.victory-conditions :as victory-conditions]
             [persistence.persistence :as persistence]
-            [api.helper :as helper]
+            [api.player-view :as player-view]
+            [api.player-view-functions :as functions]
```
kenan-rhoton: I think `api.player-view-functions` should only be used from inside `api.player-view`...

Masclins: where should I put `player-num`, then? :(

kenan-rhoton: maybe move `player-num` and `translate-player` to `api.conversions` or something?

backend/src/api/conversions.clj
```diff
@@ -0,0 +1,23 @@
...
+  [game-state internal me]
+  (cond
+    (nil? internal) nil
+    (>= internal 2) :tie
+    (= me (get-in game-state [:player-ids internal])) :me
+    :else :opponent))
+
+(defn ^:private uuid [] (str (java.util.UUID/randomUUID)))
+
+(defn create-player
```
kenan-rhoton: `generate-player-uuid`? Also, this doesn't fit as a "conversion", maybe on its own file?

PR #73, opened 2018-03-17
THE BIG FAT FRONTEND REMAKE!

kenan-rhoton:
Things changed:

- We use `webpack` to use Javascript syntax from this century
- Gone is `XmlHttpRequest`, we use `node-fetch` for  more readability
- Code has been split further
- (Controversial) We killed HTML in code, we instead use `.pug` files that generate HTML. Why? Now we can remove repeated stuff and have files that extend each other sensibly.
- As part of the cleanup and webpack use, `require` now works in a sensible way out-of-the-box and we can require relative to the `src`.
- We now have a possibility to easily add linters, tests or anything we want through webpack
- Everything can get live-reloaded on file change
- ¿Did I mention everything is more readable now?

Please! Critique this thoroughly, I want frontend to be a joy to work with, not a chore.

I especially think this pull request is useful cause it allows us to unit-test game easily if we want to (I would make a separate issue for that) and turn it into something that can be TDD'd just like backend.


kenan-rhoton:
If you like these changes, I'll add pointers in README.md and CONTRIBUTING.md to make things clear.

**ANYTHING** you don't understand, please ask.

e2e/game.js
```diff
@@ -3,74 +3,34 @@ import { Selector, ClientFunction } from 'testcafe';
...
-    
+
     await testCase
         .navigateTo(myGameURL.href)
         .expect(gamePage.cardsInHand.exists).ok()
-	.expect(gamePage.myScore.innerText).eql("0")
-	.expect(gamePage.opponentScore.innerText).eql("0");
-    
+        .expect(gamePage.myScore.innerText).eql("0")
+        .expect(gamePage.opponentScore.innerText).eql("0");
```
Masclins: What's the difference here?

e2e/game.js
```diff
@@ -3,74 +3,34 @@ import { Selector, ClientFunction } from 'testcafe';
 import MainPage from './main_page.js';
 import GamePage from './game_page.js';
```
Masclins: This file is for tests, isn't it? Yet it is not in a test-named folder nor file. Isn't it confusing?

e2e/wait-for-frontend.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
+import fetch from 'node-fetch';
+import delay from 'delay';
+
+export default async function() {
```
Masclins: What does this file do? I don't think I understand it :(

e2e/game.js
kenan-rhoton: Indentation

e2e/game.js
kenan-rhoton: It's in `e2e/` and it's for the End-to-end tests (tests that simulate a browser).

Unit tests will later go in `frontend/test`.

e2e/wait-for-frontend.js
kenan-rhoton: right now, it has an error I just realized, due to my forgetting to import `config`.

What it does is try to load the frontend website and retry one second later if it fails. It returns immediately on success. It's a way to make sure the e2e tests are never launched before frontend is running.

e2e/game.js
Masclins: Could it be called `e2e_tests` or something alike, then?

docker-compose.e2e.yml
```diff
@@ -1,9 +1,9 @@
 version: '2'
 services:
     e2e:
-        build: e2e
+        build: e2e_tests
```
Masclins: Thanks

frontend/src/game/game.pug
```diff
@@ -0,0 +1,39 @@
+extends ../tcg
```
Masclins: Why are we *really* using `.pug` for? It seems a little bit more readable, but I'm afraid it's less generic that `.html` and it doesn't seem to improve anything that much.

kenan-rhoton: mainly to have things like `extends` and `include` which allow us to keep things more organized instead of requiring all HTML to be on a single file (which could get very ugly as `src/game` gets more complex)

kenan-rhoton: There are other alternatives, but this one had a funny name ¯\\_(ツ)_/¯

Masclins: Not much ago we make a huge change, removing vue. I'm afraid we are now adding new stuff, like vue, that we might want to remove later.

kenan-rhoton: The problem with vue was mainly that we had added it before we needed it. I even made a comment on how we possibly might actually reinclude it later if it was actually needed.

These changes mainly solve two things:
1. More readable and easy to work with Javascript (especially regarding requires and fetching data from backend)
2. More readable and easy to work with HTML (we can now split files to organize them and no longer require copy-pasting a bunch of stuff!)

kenan-rhoton: Also, I am completely open to using a different alternative for either problem, although `webpack` alternatives are kinda iffy.

But I can make a list of alternatives to `pug` if you prefer 🙃

PR #79, opened 2018-03-19
Added function to modify card power by increase or decrease

### LINKED ISSUE #74: Add backend function to add or subtract power to cards
This issue is related to #41, and requires a thin layer over `alter-card`

We need a function that takes a game-state and a path to a card and can increase/decrease that card's power (or ideally execute any function on it).

kenan-rhoton:
Closes #74.

I think I can improve naming on this and make more varied tests.

### DIFF
--- a/backend/src/rules/alter_card.clj
+++ b/backend/src/rules/alter_card.clj
@@ -5,8 +5,9 @@
   [game-state path new-values]
   (update-in game-state path #(merge % new-values)))
 
-(defn increase-power
-  "Alters a cards' power, by adding the passed value to it"
+(defn add-power
+  "Alters a cards' power, by adding the passed value to it.
+  This power may be negative, resulting in a decrease"
   [game-state path increase]
   (let [power (get-in game-state (conj path :power))]
     (alter-card game-state path {:power (+ power increase)})))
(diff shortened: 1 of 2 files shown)


backend/test/rules/effects_test.clj
```diff
@@ -58,3 +58,29 @@
           (play-card/play-card 1 1 1)
           (alter-card/alter-card [:rows 0 0] {:power 9001})
           :rows (get 0) (get 0)))))
+
+(defexpect alter-card-relative
```
Masclins: I miss a test when cards have more attributes than `:power`, checking it keeps the other attributes intact.
Also, you are checking the same `path` on the three tests, that doesn't ensure it works as intended.

kenan-rhoton: Completely agreed on both accounts

backend/src/rules/alter_card.clj
```diff
@@ -5,7 +5,7 @@
   [game-state path new-values]
   (update-in game-state path #(merge % new-values)))
 
-(defn relative-power
+(defn increase-power
   "Alters a cards' power, by adding the passed value to it"
```
Masclins: I'd comment here that it can "increase" negative values (therefore decreasing it)

Issue #76, opened 2018-03-19
Change backend play-card functions to allow for an optional target


kenan-rhoton:
Because of #41, sometimes cards will be played with a target.

PR #82, opened 2018-03-20
Added optative target to play a card

Masclins:
This solves #76 and makes it able to use add a target variable when playing cards.


backend/test/rules/simultaneous_test.clj
```diff
@@ -16,12 +16,12 @@
 (defexpect next-play
   ; Stores next-play
   (expect
-    {:player 0 :index 0 :row 0}
+    {:player 0 :index 0 :row 0 :target nil}
```
kenan-rhoton: I feel like this change should'nt be needed, nonexistence of the key should be treated as if it were nil.

Masclins: yet it is stored as `nil`. I'm not sure I understand what you'd expect here.

kenan-rhoton: 🤔

Masclins:
Is there anything that should be changed?

kenan-rhoton:
Yes, my forgetfulness

PR #83, opened 2018-03-20
Rules tests rewritten

### LINKED ISSUE #80: Backend: Improve tests
I've realized many backend tests have more dependencies than needed (`effects_test` really does NOT need to call `create-game` for anything) as well as some that can simply be greatly improved in style and clarity.

It would help a lot to have another pass and maybe establish concrete guidelines for unit tests.

kenan-rhoton:
Relates to #80.

The following criteria are being followed:
- Minimize the amount of dependencies in each test file
- Have each test file be about a single unit when possible
- Explode big test files into their own folder


kenan-rhoton:
Api tests refactor is still missing.

backend/test/rules/new_game_test.clj
```diff
@@ -25,10 +26,25 @@
 
   ; Rows begin empty
   (expect
-    [[] [] [] [] []]
+    (repeat 5 [])
```
Masclins: I prefere to see what I will get in a test, but ok...

backend/test/rules/row_winner_test.clj
```diff
@@ -1,36 +1,24 @@
...
+    (victory/get-won-rows {:rows [[{:power 1 :owner 5}] [] [] [] []]} 5))
+
   (expect
-    #(= (victory/get-won-rows % 0) 0)
-    (autoplay/as-rules
-      (fn [i] 0)
-      (fn [i] 0))))
+    4
+    (victory/get-won-rows
+      {:rows [
```
Masclins: Good move there!

backend/test/rules/new_game_test.clj
kenan-rhoton: This I can see both ways, I changed one direction and back like 6-7 times 😂
I would just ask that we be consistent throughout

PR #84, opened 2018-03-20
API tests: persistence mocked, some refactoring

kenan-rhoton:
More stuff for #80.

As a bonus, it is now possible to run `lein test` and `lein test-refresh` locally within the `backend` folder, without need of docker, due to all calls to persistence being mocked in tests.

### DIFF
--- a/backend/test/api/newgame_test.clj
+++ b/backend/test/api/newgame_test.clj
@@ -1,50 +1,48 @@
 (ns api.newgame-test
   (:require [expectations.clojure.test :refer :all]
+            [clojure.test :as ctest]
+            [mocking :as mocking]
             [api.base :as api]
             [configs.hands :as hands]
             [configs.messages :as messages]))
 
+(ctest/use-fixtures :each mocking/mock-persistence)
+
 (defexpect sanity-check
-  (expect
-    #(contains? % :game-id)
-    (api/create-game))
-  (expect
-    #(contains? % :player-id)
-    (api/create-game))
-  (expect
-    #(= (count (:hand %)) (count hands/default-hand))
-    (api/create-game))
-  (expect
-    #(= 5 (count (:rows %)))
-    (api/create-game)))
+  (let [game (api/create-game)]
+    (expect
+      #(contains? % :game-id)
+      game)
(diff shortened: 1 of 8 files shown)


backend/test/api/newgame_test.clj
```diff
@@ -1,50 +1,48 @@
...
-  (expect
-    #(contains? % :player-id)
-    (api/create-game))
-  (expect
-    #(= (count (:hand %)) (count hands/default-hand))
-    (api/create-game))
-  (expect
-    #(= 5 (count (:rows %)))
-    (api/create-game)))
+  (let [game (api/create-game)]
```
Masclins: This is way prettier

backend/test/api/newgame_test.clj
```diff
@@ -1,50 +1,48 @@
...
-        (api/add-player)
-        :game-id
-        (api/add-player)))
+    ; A third player causes an error
+    (expect
+      {:error messages/too-many-players}
+      (api/add-player (:game-id game)))
 
-  (expect
-    #(not (= (:player-id %) (:player-id (api/add-player (:game-id %)))))
```
Masclins: How much I love stop using `#(` when an alternative is clearer!

backend/test/api/scores_test.clj
```diff
@@ -1,56 +1,59 @@
...
-          player-id (:player-id game-state)
-          opponent-id (:player-id (api/add-player game-id))]
+  (let [game-state (api/create-game)
+        game-id (:game-id game-state)
+        player-id (:player-id game-state)
+        opponent-id (:player-id (api/add-player game-id))]
+
+    ; Game begins with scores at 0
+    (expect
+      [[0 0] [0 0] [0 0] [0 0] [0 0]]
```
Masclins: Now we don't use `(repeat`? This is the decided guidelines for tests?

kenan-rhoton: I think yes, we should try to always have an `expect result` syntax, where it's the object expected, whatever its type or form (in some very few cases this might be terribly impossible or impractical, but they should be a minority)

Issue #88, opened 2018-03-21
Possibility to configure a row limit


Masclins:
To solve #42:

We should be able to configure a row limit for each row.
Rows should be something like: `(:rows game-state) => [{:limit 4 :cards []} {:limit 4 :cards []} {:limit 4 :cards []} {:limit 4 :cards []} {:limit 4 :cards []}]`, with that default limit found in `configs/rows.clj`

kenan-rhoton:
This should be set in `configs/config.yml` (from the root) as the first issue to start this process, I think

Masclins:
Why it wasn't automatically closed, now?

kenan-rhoton:
"Solves" is not one of the keywords.

The pull request message has to contain one of these: `closes #1`, `fixes #1` or `resolves #1`.

https://help.github.com/articles/closing-issues-using-keywords/

Issue #90, opened 2018-03-21
Avoiding plays beyond row limit


Masclins:
To solve #42:

Backend should not allow a player to play a card in a row in which there are equal or more than its limit cards owned by him.

Masclins:
Right now this already happens without need to change anything.
Maybe it would be interesting to add some kind of visual notification.
Even though, I'm not sure if it is really needed for release v0.2.0

kenan-rhoton:
Would defer this for another release.

PR #96, opened 2018-03-22
Limit can be configured

Masclins:
This solves #88.

It required changing the structure of `:rows` and adapting the backend tests for it. 

Right now limit configured is used for nothing, and the rows are passed to frontend in the same format as before.

### DIFF
--- a/backend/src/api/base.clj
+++ b/backend/src/api/base.clj
@@ -13,11 +13,11 @@
   (player-view/get-game-as-player (persistence/fetch-game game-id) player-id))
 
 (defn play-card-as-player
-  [game-id player index row & target]
+  [game-id player index row-id & target]
   (let [game-state (persistence/fetch-game game-id)]
       (if (= (:status (get-game game-id player)) messages/play)
           (do
-            (persistence/save-game (play-card/play-card game-state (conversions/player-num game-state player) index row (first target)))
+            (persistence/save-game (play-card/play-card game-state (conversions/player-num game-state player) index row-id (first target)))
             (get-game game-id player))
           {:error messages/out-of-turn})))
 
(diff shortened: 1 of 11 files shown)


backend/src/api/base.clj
```diff
@@ -13,11 +13,11 @@
   (player-view/get-game-as-player (persistence/fetch-game game-id) player-id))
 
 (defn play-card-as-player
-  [game-id player index row & target]
+  [game-id player index row-id & target]
```
kenan-rhoton: yay, clarity!

backend/src/api/player_view_functions.clj
```diff
@@ -8,13 +8,19 @@
...
+
 (defn get-rows
   "Return the rows as seen by the player"
   [game-state player-id]
-  (mapv #(mapv (fn [card]
-                 (assoc card :owner
-                   (conversions/translate-player game-state (:owner card) player-id)))
-               %)
+  (mapv (fn [row]
+          (get-row-cards (:cards row) game-state player-id))
```
kenan-rhoton: You are currently eliminating in this step any information other than the contents of `:cards` from a row. Is this what you intend?

Masclins: It is **right now**.
My only intent with this pull-request is to change the backend so it handles row limits, but frontend will be done in another pull-request (for #90).
So the api should keep giving the same as before to frontend. That's why here it loses information.

backend/test/rules/play/next_play.clj
```diff
@@ -2,18 +2,21 @@
   (:require [expectations.clojure.test :refer :all]
             [rules.play-card :as play-card]))
 
+(def empty-rows
+  [{:cards []} {:cards []} {:cards []}])
```
kenan-rhoton: I definitely like this.

backend/src/api/player_view_functions.clj
kenan-rhoton: I thought so, but wanted to double-check and comment so we don't forget 🙃

PR #99, opened 2018-03-23
Back-end handles row limit

### LINKED ISSUE #89: Backend using row limits
To solve #42:

Backend should not allow a player to play a card in a row in which there are equal or more than `:limit` cards owned by him.

Upon playing a card there an error should be returned.

### LINKED ISSUE #100: Allow row-limit to be configured differently for each row
Right now limit can only be set as one value, equal to each row.
`create-game/new-game` should be changed so it allows a different limit on each row, and this to be passed from `configs.rows`.

Masclins:
This resolves #89 and resolves #100 (edited).

The default value for the limit is set to `9000` while frontend cannot handle limits.
Also, that value is used for limitless cases.


Masclins:
With the latter commit it also resolves #100

kenan-rhoton:
Shouldn't this be defined in a `config/settings.yml` file?

Masclins:
`configs/settings.yml` would be public, wouldn't it?
Right now hand and limits can be public, since it's always the same, but when we have different hands and limits for each game, I really doubt we want that as public.
I feel like it's better to have game configurations only on the backend.

kenan-rhoton:
I was halfway through writing a counterargument when I found a much greater counterargument to my own counterargument: We will send frontend the limit information **inside** the rows, anyway.

Counterargument. (I like that word)

backend/src/api/base.clj
```diff
@@ -23,7 +23,8 @@
 
 (defn add-player
   "Adds a player to a game"
-  [game-id]
+  ([game-id] (add-player game-id {}))
```
kenan-rhoton: Ok,`add-player` is doing too many things already.

`create-game` should be the one creating a (playerless) game with its initial config, and then call add-player to simply to that: add a player. This would greatly simplify these functions (and I once again know its not specifically related to this issue, I might just open a separate issue now)

backend/src/rules/create_game.clj
```diff
@@ -11,16 +11,13 @@
...
   ([ini-config]
    {
-    :players (vec (repeat 2 (new-player (if (contains? ini-config :hand)
-                                            (:hand ini-config)
-                                            hands/default-hand))))
-    :rows (vec (repeat 5 {:limit (if (contains? ini-config :limit)
-                                     (:limit ini-config)
-                                     rows/default-limit)
-                          :cards []}))
+    :players (vec (repeat 2 (new-player (:hand ini-config hands/default-hand))))
```
kenan-rhoton: great use of the default value option in `(:key map default)` 🙃

backend/src/rules/create_game.clj
```diff
@@ -11,16 +11,13 @@
...
+    :rows (vec (reduce
+                 #(concat %1 [{:limit %2 :cards []}])
+                 []
+                 (:limits ini-config rows/default-limit)))
```
kenan-rhoton: Readability nitpick: `rows/default-limit` reads like you expect and integer and thus the `reduce` makes no sense. `rows/default-limits`(with an s) would make it clear that it's a collection

backend/src/rules/play_card.clj
```diff
@@ -26,13 +26,13 @@
 (defn ^:private apply-play-card
   "Plays a card waiting to be played onto the board"
   [game-state play]
-    (let [player (:player play)
+    (let [player-id (:player play)
```
kenan-rhoton: How do we *still* have so many placer where we forgot to put `-id`?

backend/src/rules/create_game.clj
Masclins: True. If you see previous commits you'll see I changed how limit configuration works. I changed `:limit` for `:limits` but forgot to do the same with the configs. Gonna do.

backend/src/rules/play_card.clj
Masclins: I keep changing them whenever I see them.

backend/test/autoplay.clj
```diff
@@ -1,12 +1,13 @@
...
   (:require [configs.hands :as hands]
+            [configs.rows :as rows]
             [api.base :as api]
             [rules.create-game :as create-game]
             [rules.play-card :as play-card]))
 
 (defn as-api
   [strategy1 strategy2]
-  (let [game (api/create-game)
+  (let [game (api/create-game {:limits rows/limitless})
```
kenan-rhoton: wait, isn't `row/limitless` the default? I still like it for readability, but now I question if I've understood

backend/src/rules/play_card.clj
kenan-rhoton: DOING THE LORD'S WORK, SON

backend/test/autoplay.clj
Masclins: Right now `rows/limitless` and `rows/default` are exactly the same. `rows/default-limits` can be changed without any problem (within backend). When frontend handles the limits, we will be able to change `rows/default-limits` without having to touch any test.

Issue #91, opened 2018-03-21
Showing row limit


Masclins:
To solve #42:

Frontend should show the row limit for every row.

PR #102, opened 2018-03-26
Frontend recieving limit

Masclins:
Small step for #90 and #91.

Now Frontend recieves the rows as backend handles them.


Masclins:
Outdated by #103

Issue #98, opened 2018-03-23
Backend test cleanup 2: Electric Boogaloo


kenan-rhoton:
So I've been giving this a lot of thought and... we're "overtesting".

What I mean is, we have tests that check our rules (in `tests/rules`) and then we have tests for our api which don't really check the api but check the internal functions and make sure they "fulfil the rules".

This is bad, we're testing the same things twice and that is why we have "duplicated" tests within `tests/rules` and `tests/api`.

The rules tests are well done, but the api tests should actually:

1. **USE** the API (that means call into `handler/entry` with an URL and a GET, POST or whatever request)
2. Just test that the calls go as expected and that they communicate with the game rules.

How do we do this, well, for step 1 I've started doing this:

```clojure
(defn request [resource method & params]
  (json/parse-string
    (:body 
      (handler/entry {:request-method method
                :uri resource
                :params (first params)}))
    true))
```

Which can be called by `(request "/games" :post)` for example to issue a "Create Game" action.

For step 2, we would need to mock out the rules we expect to be called and then `expect` them to have been called in the way they should.

This would greatly reduce the amount of tests in API, while actually *increasing* our coverage (which I'm going to also add metrics for).

Thoughts? 🙃

Masclins:
Thoughts:

1. We keep adding new Issues to Release v0.1.1 only because one Issue is taking way longer than expected. If that Issue were resolved on time (being *on time* when it was said to be resolved) there's no way this Issue got into Release v0.1.1. Aren't we abusing of that milestone? 

2. Backend tests need a cleanup for sure, maybe even a restructure. It's hard to find where a function is tested and to decide where to nest a new test.

3. Most of the tested `api` functions do much more than simply calling a `rules` one. Could you give an example of something checked twice due to a `rules` and `api` test?

4. Still, though, it seems like a good idea to simplify most of the `api` tests by checking their answer to a request. In the end, that's what we want.

kenan-rhoton:
1. Agree
2. Agree
3. I'll give an example in short
4. Agree

kenan-rhoton:
An abstract example: everything that uses `autoplay/as-api` is actually simulating a game and thus, the rules. A more simple approach would be to mock a finished game (one without hands) and simply make the call for the status and check that the scores or whatever are correct.

A concrete (and different) example:

```clojure

(defexpect playing-cards
  (let [game (api/create-game)
        game-id (:game-id game)
        player-id (:player-id game)
[...]

PR #103, opened 2018-03-27
Add-Power on Backend

### LINKED ISSUE #75: Allow cards to modify other cards on being played
Due to #41 we want some cards to have an effect when they are played, in particular the effect of modifying other cards.

Masclins:
Allows cards to have the `:add-power` attribute, modifying cards upon play (if played with a `:target`).
I thought about creating an attribute called `:abilities` and putting all its abilities inside it. Right now we only have one ability so it seemed unpractical.

This resolves #75 and outdates #102 (since I needed those commits to check the `api` without doing the work twice)

### DIFF
--- a/backend/src/rules/play_card.clj
+++ b/backend/src/rules/play_card.clj
@@ -1,5 +1,6 @@
 (ns rules.play-card
-  (:require [configs.messages :as messages]))
+  (:require [configs.messages :as messages]
+            [rules.alter-card :as alter-card]))
 
 (defn ^:private add-card-to-row
   "Adds a card onto the specified row"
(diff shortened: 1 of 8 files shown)


backend/test/api/effects_test.clj
```diff
@@ -0,0 +1,23 @@
+(ns api.effects-test
+  (:require [expectations.clojure.test :refer :all]
+            [clojure.test :as ctest]
+            [mocking :as mocking]
+            [api.base :as api]))
+
+(ctest/use-fixtures :each mocking/mock-persistence)
+
+(defexpect playing-effects
```
Masclins: I'm not sure this test is needed, but better to be safe than sorry.

backend/src/rules/play_card.clj
```diff
@@ -23,14 +24,26 @@
...
+  (if (nil? (:target play))
+    game-state
+    (alter-card/add-power
+      game-state 
+      (:target play)
+      (get-in game-state [:players (:player play)
+                          :hand (:index play)
+                          :add-power]))))
+
+(defn apply-play-card
```
Masclins: No longer private so we can call it from tests and make them more targeted to what we wanna test.

Masclins:
I thought about another alternative, but I feel like it doesn't have much sense now that we only have one ability:

- *cards* have `:abilities {:add-power x :destroy-card true}` (`true` or `nil` would be used for abilities that require no value. Also could be used for adding future restrictions there)

- *next-play* have `:targets {:add-power target :ability-n2 target}`

backend/test/rules/effects_test.clj
```diff
@@ -33,3 +34,12 @@
       [nil {:croissant "chocolate" :enchilada {:power 1 :salsa "mild"}}]
       [1 :enchilada]
       55)))
+
+(defexpect playing-add-power
```
kenan-rhoton: Would like a second test with different values here 🙃

backend/test/api/effects_test.clj
kenan-rhoton: The test is fine for now, although it's the kind of test I'm wanting to simplify with #98

PR #104, opened 2018-03-27
cleaner api/create-game

### LINKED ISSUE #101: Define create-game and add-player correctly
`add-player` is doing way too many things, and some of them should be done by `create-game`.

- `create-game`: Initialize a player-less game, with default configs.
- `add-player`: add a player to a game.
- `create-game-with-player`: Call the first two functions in order, this will be the new entry point for game creation in the API.

Masclins:
Resolves #101 though not with the exact names as in the issue.

- `create-empty-game`: Initialize a player-less game, with default configs.
- `add-player`: add a player to a game.
- `create-game`: Call the first two functions in order, this will be the new entry point for game creation in the API.

This way, only `api/base.clj` had to be changed.

### DIFF
--- a/backend/src/api/base.clj
+++ b/backend/src/api/base.clj
@@ -21,22 +21,29 @@
             (get-game game-id player))
           {:error messages/out-of-turn})))
 
+(defn create-empty-game
+  "Creates a new instance of a game"
+  [& ini-config]
+  (let [game-id (persistence/next-id)]
+    (persistence/save-game
+      (assoc (create-game/new-game (first ini-config))
+             :game-id game-id))
+    game-id))
+
 (defn add-player
   "Adds a player to a game"
-  ([game-id] (add-player game-id {}))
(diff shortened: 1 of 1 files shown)


pernilsalat:
I think it is important that a function accepts only what needs in order to, well, function

```
(defn create-empty-game
  "Creates a new instance of a game"
  [& ini-config]
  (let [game-id (persistence/next-id)]
    (persistence/save-game
      (assoc (create-game/new-game (first ini-config))
             :game-id game-id))
    game-id))

(defn create-game
  "Creates a new instance of a game with a player"
  [& ini-config]
  (-> (create-empty-game (first ini-config))
      (add-player)))
```
In this case, `create-empty-game` accepts everything while it only needs the first argument given, and actually it is only called with one argument. It is a better approach to contemplate the case when it is called without parameters (if that is something that **we want**) and with the one needed rather than accepting everything.

PR #106, opened 2018-03-28
New Default Configuration

### LINKED ISSUE #85: Default configuration (hand and row-limit)
v0.2.0 should be released with the following default configuration:
- Hand: 10, 9, 8, 7, 6+1, 6-1, 4+2, 4-2 (+/- x means it changes another card's value).
- Row-limit: 4 on each row.

This probably won't be doable until #41 and #42 are solved.

Masclins:
It seems like, considering how everything is done, the new configuration doesn't invalidates anything else.
Therefore, it resolves #85.

Still, frontend is ignoring the row limits and add-power abilities.

I included a new hand in `config/hands.clj` for `autoplay`. Even though it works, now, it is a bad idea that it plays cards with abilities without using them.

### DIFF
--- a/backend/src/configs/hands.clj
+++ b/backend/src/configs/hands.clj
@@ -2,15 +2,11 @@
 
 (def default-hand
   "Default hand (can't be empty)"
-  [{:power 5}
-   {:power 6}
-   {:power 7}
-   {:power 8}
+  [{:power 10}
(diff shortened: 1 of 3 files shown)


Masclins:
Maybe there's no need to remove autoplay, using "simple-hand"

Issue #55, opened 2018-03-07
Cards in hand identified by independent index


Masclins:
We are changing almost everything (mostly backend) so we have an array of all the cards (therefore, identified by an independent index). Cards in this array will have an attribute showing the zone in which the card is. That means a change from "Zones with cards" to "Cards in zones".

Everything from `backend/src/rules` is already changed, including tests.
`backend/src/api` needs to have the tests changed so they use the format "one-test-file-per-src-file".
Now the following needs to be done:

- [x] Double-check backend tests (and add any that's needed/missing)

- [x] Tests for `api/base.clj`: Put the current ones together.

- [x] Tests for `api/base.clj`: Test `(play-card-as-player`

- [x] Tests for `api/conversions.clj`, `api/generators.clj` and `api/handler.clj`. This should be pretty easy, since most of those don't do much.

- [x] Adapt all tests on `backend/old_tests/` (this should be covered by the previous two points, but just in case...)

kenan-rhoton:
Giving cards an `id` field?

Masclins:
Yes. 

And I assume it would be best if it's different for every card (so no shared `id` between my cards and my opponent's)

kenan-rhoton:
Should be assigned on game creation then

Masclins:
I'd like to hear from @pernilsalat and what advantages this would have, before begining working on it.

kenan-rhoton:
Right now zero, it will probably be useful in the future (especially if we want to allow a player to reorganize cards in his hand in frontend). Wait, **I** assigned this to v0.1.0? Woah, I must've been drunk.

Masclins:
Thinking about Issue #41 it would probably be useful to have cards identified by id, also in the board.

kenan-rhoton:
Yes, as card move zones they should keep their id.

Masclins:
It seems like we are working on targets before giving id to cards. 
Do we really want `:target` to be a path, rather than an `:id`?

kenan-rhoton:
I know I don't 🙃

Masclins:
`:target` has been implemented to be a path (shouldn't be hard to change).

The thing is... Knowing the `:id` of a card is easy, while knowing it's path is not that easy (specially when frontend will have to tell backend its path).
On the other hand, finding where a card that has to be changed, based on `:id` might be hard, while doing so with path is easy.

[...]

PR #108, opened 2018-04-03
First part for Indexes (+ other Issues)

### LINKED ISSUE #75: Allow cards to modify other cards on being played

### LINKED ISSUE #87: Different hands for each player
Game should allow players to have different initial hands, so no player has perfect information.

Masclins:
This resolves #75, resolves #87 and is the first part for #55.

I merged the pull-request #103 and #105 into this for being able to work on #55.

### DIFF
--- a/backend/src/rules/create_game.clj
+++ b/backend/src/rules/create_game.clj
@@ -2,11 +2,20 @@
   (:require [configs.hands :as hands]
             [configs.rows :as rows]))
 
-(defn ^:private new-player
+(defn new-player
   "Creates a new player object"
-  ([hand]
+  ([hand player]
    {
-    :hand hand
+    :hand (loop [final-hand []
+                 base-hand hand
+                 id (* 1000 player)]
+            (if (empty? base-hand)
+              final-hand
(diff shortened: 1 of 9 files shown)


Masclins:
...I think I messed up, but don't know where :'(

backend/src/rules/create_game.clj
```diff
@@ -2,22 +2,35 @@
...
-(defn ^:private new-player
+(defn new-player
   "Creates a new player object"
-  ([hand]
-   {
-    :hand hand
-    }))
+[hand player]
+{
+ :hand (loop [final-hand []
```
kenan-rhoton:
```clojure
(mapv (fn [card id] (assoc card :id id)) hand (iterate inc 1000))
```

Masclins: I don't think this does the same as the loop.
The loop is supposed to use indexes begining with 0 for player 0, and beginning with 1000 for player 1.

PR #112, opened 2018-04-16
showing add-power

### LINKED ISSUE #107: Cards in hand show abilities
Right now front-end only shows the power of cards. They also should show their abilities.

Given we only have the add-power ability, maybe we could do something like `power (add-power)` which will end up as `6 (+1)` (the '+' would be nice).

Masclins:
This resolves #107

### DIFF
--- a/frontend/src/game/hand.js
+++ b/frontend/src/game/hand.js
@@ -12,7 +12,13 @@ module.exports = {
 
         handState.forEach(function (cardInHand, index) {
             var newCard = templates.baseCard.cloneNode(true);
-            newCard.innerHTML = cardInHand["power"];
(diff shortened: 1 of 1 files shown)


frontend/src/game/hand.js
```diff
@@ -12,7 +12,13 @@ module.exports = {
 
         handState.forEach(function (cardInHand, index) {
             var newCard = templates.baseCard.cloneNode(true);
-            newCard.innerHTML = cardInHand["power"];
+            newCard.innerHTML = cardInHand["power"]
+            var addpower = cardInHand["add-power"]
+            if (addpower != undefined) {
```
kenan-rhoton: I would like this extracted into a function

Issue #78, opened 2018-03-19
Communication of playing a card with targets


kenan-rhoton:
For #41.

This requires there be an E2E test in place, and that the feature works as expected. This is the final step in implementation of issue #41.

Masclins:
This probably has more sense to be worked with after resolving #55

Masclins:
...Now we have the problem of how to identify a card in the frontend's board with a card in the backend's board. This seems complicated because they are sorted differently, since we split them by owner on frontend.

kenan-rhoton:
This might be the first actual case for an ID... :thinking: 

We should discuss options.

Masclins:
This is actually the Issue that @pernilsalat said would work this next two weeks.

PR #115, opened 2018-04-18
Making cards with ability require a target-selection

### LINKED ISSUE #77: Player can select a target when playing a card with abilities
Due to #41, this is the frontend presentation issue.

Masclins:
This is the first commit on an attempt to resolve #77.

As I understand, the only thing that should be achieved to solve this Issue is the game requiring a target to be selected when playing a card with an ability. Doing something with that target will be solved by #78.


Masclins:
This commit is **really** wrong, but is the first approach I managed to do for solving the issue.

1. It uses a lot the `status.clickedCard`, to the point that dragging cards now give an error.
2. It can't click owned cards. It seems the problem is that clicking on an owned card implies also clicking on an owned row.

Any help with this two problems will be appreciated.

Masclins:
Now the only problem is making the card green when clicking an owned card.

...and, of course, the tests. All the tests...

frontend/src/game/board/builder.js
```diff
@@ -40,6 +42,12 @@ module.exports = {
         newCard.classList.remove("col-1");
         newCard.classList.add("col-2");
         newCard.innerHTML = cardData["power"];
+        newCard.addEventListener('click', function() {
+            if (status.clickedCard !== undefined &&
```
pernilsalat: `if (status.clickedCard !== undefined)   `
is equivalent to 
` if (status.clickedCard)`

frontend/src/game/board/builder.js
```diff
@@ -40,6 +42,12 @@ module.exports = {
...
+                status.clickedCard.hasAttribute("row-played")) {
+                play.playCard(true)
```
pernilsalat: wut

Masclins: Si apreto una carta en joc i ja tinc una carta seleccionada i jugada (és a dir, l'hi he assignat quina és la seva `row-played` jugo la `status.clickedCard`. 
El `true` ara per ara és per indicar que ja s'ha seleccionat el target. Podria ser qualsevol valor que no sigui `undefined`. Quan fem la issue que passa el target, aquest valor serà el target.

frontend/src/game/board/builder.js
Masclins: Thanks.

frontend/src/game/board/builder.js
pernilsalat: then i feel this `true` should be in a variable, and that variable would be set to true for now

frontend/src/game/hand.js
```diff
@@ -23,8 +23,9 @@ module.exports = {
 
         handState.forEach(function (cardInHand, index) {
             var newCard = templates.baseCard.cloneNode(true);
-            newCard.innerHTML = cardInHand["power"]
-            newCard.innerHTML += showAddPower(cardInHand["add-power"])
+            newCard.innerHTML = cardInHand["power"];
+            newCard.setAttribute("add-power", cardInHand["add-power"]);
+            newCard.innerHTML += showAddPower(cardInHand["add-power"]);
```
Masclins: The ability `add-power` is set as an attribute, so we can easily check it for a card.
Maybe this two lines can be simplified.

frontend/src/game/play/play.js
```diff
@@ -7,23 +7,27 @@ const config = require('config/config.js');
...
 
 module.exports = {
-    playCard(rownum, cardindex, target) {
-        const playData = {
-            index: cardindex,
-            row: rownum,
-            target: target,
-        };
-        document.querySelector('.card[index="'+cardindex+'"]').style.background = "yellow";
+    playCard(target) {
```
Masclins: Since we keep track of the `status.clickedCard` for marking it and handling abilities, passing the playing info here is redundant.

frontend/src/game/play/play.js
```diff
@@ -7,23 +7,27 @@ const config = require('config/config.js');
...
+        if (target === undefined &&
+            status.clickedCard.getAttribute("add-power")) {
+            status.clickedCard.style.background = "yellow";    
+        } else {
```
Masclins: I go ahead with the playing only if the card doesn't have `add-power` or has a defined `target`. Otherwise I simply mark the card in yellow.

frontend/src/game/play/play.js
```diff
@@ -7,23 +7,27 @@ const config = require('config/config.js');
...
 
 module.exports = {
-    playCard(rownum, cardindex, target) {
-        const playData = {
-            index: cardindex,
-            row: rownum,
-            target: target,
-        };
-        document.querySelector('.card[index="'+cardindex+'"]').style.background = "yellow";
+    playCard() {
```
kenan-rhoton: I don't like this function "magically" getting its info from somewhere that isn't even in the same file (`status.clickedCard`).

It should take the clicked card as an argument or it will grow unmantainable in the future, as this project grows.

kenan-rhoton:
We need a test that checks that on clicking to play a card with an ability, a target selection is requested.

kenan-rhoton:
Unit tests are failing 🙃

PR #117, opened 2018-04-19
Centering Row Scores

### LINKED ISSUE #111: Centering scores
board

EDIT:
**Opponent's card's** was solved with a minor change.

The row-scores should be centered. Right now the innerHTML there seems to ignore any attempt to center them.

Masclins:
By simply removing some classes of it, resolves #111


Masclins:
This might resolve #117 but I'm not 100%...

Masclins:
Don't get why it fails... :S

PR #119, opened 2018-04-25
Frontend passes target to backend

### LINKED ISSUE #78: Communication of playing a card with targets

Masclins:
This resolves #78.

But after this and #115 everything is **really** ugly.

Philosophy goes as following:
Cards in hand have an attribute `add-power` equal to its *add-power* ability or `"undefined"`.
When a card is "played", if `add-power === "undefined"` the play is passed to the backend, with `target: "nil"`. Otherwise it gets an attribute `row-played` with the row where it's been played.
When a card on the board is clicked with a card "played" (that means it must be a card with an `add-power !== "undefined"` previously dragged/clicked) the "played" card recieves an attribute `target: "[:rows 0 :cards 0]"` with 0's being the appropriate indexes (even I can see the ugliness in that).

On the other hand, backend recieves always a target, and `api.base/play-card-as-player` understand the target (a string) as a clojure code giving the path (using `(let [target (load-string target)]`.

From this pull-request (and the previous one) I see a couple of things that need work:
- We should decide how to handle targets. I coded everything regarding targets and most probably are suboptimal (I'm being really generous here).
- We are leaving a lot of *work* to the frontend, like deciding which cards need a target.
- Attribute-setting and calls to `play()` are done in many places on the frontend code. That's weird and messy.


backend/src/api/base.clj
```diff
@@ -13,13 +13,16 @@
   (player-view/get-game-as-player (persistence/fetch-game game-id) player-id))
 
 (defn play-card-as-player
-  [game-id player index row-id & target]
-  (let [game-state (persistence/fetch-game game-id)]
+ ([game-id player index row-id]
+  (play-card-as-player game-id player index row-id "nil"))
+ ([game-id player index row-id target]
+  (let [game-state (persistence/fetch-game game-id)
+        target (load-string target)]
```
kenan-rhoton: `load-string` is a security issue, it is executing arbitrary Clojure code contained in a string received from an external connection.
A malicious user could alter the value being sent as `target` and run any kind of code on the system, compromising the entire backend.

frontend/tests/game/play.test.js
```diff
@@ -12,11 +12,12 @@ test('Playing cards changes background as intended', () => {
     play.playCard(card);
     expect(card.style.background).toBe("green");
     
+    card = document.createElement("div");
     card.setAttribute("add-power", 1);
     play.playCard(card);
     expect(card.style.background).toBe("yellow");
 
-    card.setAttribute("target", 1);
+    card.setAttribute("target", "[:rows 0 :cards 0]");
```
kenan-rhoton: We cannot send Clojure code. Either split the information into `target-row` and `target-card` or use a json-encoded object

kenan-rhoton (changes requested):

Masclins:
Now it passes the backend and frontend tests, but not the e2e.
So I think it's pretty obvious that the problem comes with the REST api.
Any help?

backend/src/api/base.clj
kenan-rhoton: Just as a reminder: this pull request *cannot* be accepted until it doesn't use vulnerable things such as `load-string`

frontend/tests/game/play.test.js
kenan-rhoton: The removal of `load-string` makes this change now incorrect, right?

Masclins: now it should pass the card-id

Masclins: This test only checks the card has a target, it doesn't checks if the target is correct.

kenan-rhoton:
@Masclins Just to confirm, you dropped this on purpose, right?

Masclins:
I dropped this because branch was deleted and pushed unto your repository.
We can either merge the branch and @pernilsalat makes his changes on a new branch, or we wait until he improves this one.

PR #121, opened 2018-04-27
Api mocking for great justice

kenan-rhoton:
The idea is to make it so we can mock (supplant with our own code) the database as close as possible to it, to be able to actually test all of our backend code.

This mostly works.

As a bonus, I've added a `lein cloverage` action which indicates how much of our code we're covering with our tests (the idea would be to reach 100% if possible, but to not obsess over it).


kenan-rhoton:
This is already implemented in recent changes to the tests

Masclins:
Shouldn't we delete the branch?

kenan-rhoton:
Done

PR #122, opened 2018-04-30
From having "zones with cards" to "cards with zones"

Masclins:
This is the beginning of what will be a huge merge.

The idea behind this is to change the fact that on the backend we had `:players` with `:hand` and `:rows`, both with cards on it. Now we will have a vector `:cards` with all cards with it, which will have the key `:location []` telling us where it is.

This will make it easier to use and pass targets of abilities, which was proving to be pretty complicated.


Masclins:
On this first pull-request only the `backend/src/rules/` had been really changed. We (should) have all tests for that part.
Now I begun working with `backend/src/api` but a problem appear for which I'd need some help.
For making everything simpler, `game-state` has `:player-ids` with the id's of both players. That makes it really messy to create a game with a single player. I actually like not allowing to create a game with a single player, since game-rules don't allow that. Problem is that we'll need to change the api's game-creation logic, and that affects `persistence` which I don't truly understand.
@kenan-rhoton could you please help me with (or do) that specific part? Idea is that game is created when both players joined, and meanwhile we only have the first player's id and nothing else.

kenan-rhoton:
I can take care of persistence and game creation, since I have acces to your repository anyway :upside_down_face: 

These changes look awesome, by the way! :smiley:

backend/src/api/base.clj
```diff
@@ -13,38 +13,44 @@
...
-                           saved-game))]
-        (get-game game-id (last (:player-ids game-state)))))))
+  (let [lobby (persistence/fetch-game game-id)
+        players-connected (count (:player-ids lobby))]
+    (cond
+      (> players-connected 1) {:error messages/too-many-players}
+      (< players-connected 1) {:error messages/lobby-not-created}
+      :else (let [second-uuid (generators/player-uuid lobby)]
+              (-> (create-game/new-game
+                    {:player-ids [(first (:player-ids lobby))
```
kenan-rhoton, line 43: Almost went crazy during the weekend til I realized that I needed to pass the player-ids to the `new-game` constructor (which is *perfectly logical, I might add*) instead of putting them by `assoc` post-construction as with the `game-id`. In fact, I think the `game-id`should also be passed in and added to the game in `new-game` to ease the logic (everything is done at construction instead of modifying stuff afterwards)

backend/test/api/join_game_test.clj
```diff
@@ -0,0 +1,34 @@
...
+            [mocking :as mocking]
+            [api.base :as base]
+            [configs.hands :as hands]
+            [configs.messages :as messages]))
+
+(ctest/use-fixtures :each mocking/mock-persistence)
+
+(defexpect join-game
+
+  (let [joined-game (-> (base/create-game) :game-id base/add-player)]
```
Masclins, line 13: Do we need to use `(->`? Wouldn't it be more readable not using it?

Masclins (commented):
Tests seem great, just got a little comment/doubt.

PR #127, opened 2018-08-31
Travis does not run E2E tests for develop

kenan-rhoton:
Travis now should only run the e2e tests and require them to pass for merges to master.


Masclins (approved):
Easy to understand, and seems to work.

PR #129, opened 2018-08-31
Tests for api

Masclins:
Joining `api/base.clj` tests and removing already considered tests.


backend/src/rules/count_cards.clj
```diff
@@ -1,7 +1,7 @@
 (ns rules.count-cards)
 
 (defn count-cards
-  "Count all cards that has keys as in condition an sums their summand"
+  "Count all cards that has keys as in condition (or sums their summand)"
```
kenan-rhoton: *have* keys 🙃

backend/test/api/base_test.clj
```diff
@@ -1,13 +1,18 @@
-(ns api.join-game-test
+(ns api.base-test
   (:require [expectations.clojure.test :refer :all]
             [clojure.test :as ctest]
             [mocking :as mocking]
             [api.base :as base]
-            [configs.hands :as hands]
```
kenan-rhoton: Wait, this is used below, right? as `hands/default-hand`?

kenan-rhoton (dismissed):

backend/test/api/base_test.clj
Masclins: Probably that's why it fails the tests (sorry, couldn't run them).
Still... I don't understand how I removed this line... Thanks.

kenan-rhoton (approved):
This is just a cleanup + reorganization, so not much to review. Approved!

PR #133, opened 2018-09-03
play-card-as-player test

### LINKED ISSUE #97: Backend api should give descriptive errors
Backend should return messages with errors and their causes, such as "invalid id" or "game does not exist".

Especially bad are the cases where an invalid format is used, as an exception is thrown (which is incredibly ugly) and I'm not even sure what the response is in those cases.

Errors to implement:

- [x] A player plays a card that's not on her hand.
- [x] A player makes a play involving a card that doesn't exist.
- [x] A player makes a play involving a row that doesn't exist.
- [x] Player-id or Game-id doesn't exist.

Masclins:
This should complete the tests of current `api/base.clj`, solving a point from #55 

Resolves #97 (at least the needed messages we came up with)


kenan-rhoton (changes requested):
I think the missing API error messages should be included here in this commit (at least the API generated ones, Rules generated ones may go separately)

kenan-rhoton (commented):
A success test for two cards correctly played by both players is missing, right?

kenan-rhoton:
Also invalid player-id/game-id

backend/src/api/base.clj
```diff
@@ -13,12 +13,16 @@
   (player-view/get-game-as-player (persistence/fetch-game game-id) player-id))
 
 (defn play-card-as-player
+  "Plays a card give and returns the game sate as seen by the player"
   [game-id player-id card-id row-id & target]
   (let [game-state (persistence/fetch-game game-id)]
       (if (= (:status (get-game game-id player-id)) messages/play)
```
kenan-rhoton: What happens if `game-state` is `nil`? (Not in the database)

backend/test/api/base_test.clj
```diff
@@ -38,3 +38,37 @@
...
+
+(defexpect play-card-messages
+
+  (let [game (base/create-game)
+        game-id (:game-id game)
+        p1 (:player-id game)
+        p2 (:player-id (base/add-player game-id))]
+    
+    (expect {:error messages/invalid-id}
+            (base/play-card-as-player 9999999999 p1 0 0))
```
Masclins, line 50: @kenan-rhoton apparently "mocking" is making this test not as good as it should be

backend/src/api/base.clj
```diff
@@ -13,13 +13,21 @@
   (player-view/get-game-as-player (persistence/fetch-game game-id) player-id))
 
 (defn play-card-as-player
+  "Plays a card give and returns the game sate as seen by the player"
   [game-id player-id card-id row-id & target]
   (let [game-state (persistence/fetch-game game-id)]
+    (if (and (= game-id (:game-id game-state))
+          (or (= player-id (first (:player-ids game-state)))
+            (= player-id (second (:player-ids game-state)))))
```
Masclins, line 21: I'm pretty sure this can be done prettier... But how?

kenan-rhoton: I'll open an issue to clean this up cause it would involve changes to mocking, so I dont forget

kenan-rhoton (approved):

PR #136, opened 2018-09-06
Removed Legacy conversions

kenan-rhoton:
This was just legacy code... So I removed it!


Masclins (approved):
doublechecked there were no more cases. So it seems.

PR #137, opened 2018-09-07
Improved Mocking to more closely mimic reality

### LINKED ISSUE #135: Smarter mocking
Right now mocking is very very dumb (which is generally a good thing). However, it should at least simulate saving and fetching **by game ID** (which it currently doesn't), which leads to ugly hacks like in #133.

This would be fixed by the mock atom being a hash-map instead of a value

kenan-rhoton:
This resolves #135.

Mocking now stores multiple games so we can unit test just as we would with the database but without firing up Redis.

This also helped detect 2 badly-defined tests and even caught a deadly bug in `add-player`!

### DIFF
--- a/backend/src/api/base.clj
+++ b/backend/src/api/base.clj
@@ -51,6 +51,7 @@
                                   second-uuid]})
                   (assoc :game-id (:game-id lobby))
                   persistence/save-game
+                  :game-id
                   (get-game second-uuid))))))
 
 (defn create-game
--- a/backend/test/api/base_test.clj
+++ b/backend/test/api/base_test.clj
@@ -9,7 +9,7 @@
 (ctest/use-fixtures :each mocking/mock-persistence)
 
 (defexpect game-creation
-  (expect 0 (:game-id (base/create-game)))
+  (expect number? (:game-id (base/create-game)))
   (expect messages/no-opp (:status (base/create-game)))
   (expect nil (:player-ids (base/create-game)))
   (expect true (contains? (base/create-game) :player-id)))
@@ -17,7 +17,7 @@
 (defexpect join-game
 
   (let [joined-game (-> (base/create-game) :game-id base/add-player)]
-    (expect 0 (:game-id joined-game))
+    (expect number? (:game-id joined-game))
     (expect (count hands/default-hand)
             (count (filter
                      #(= "me" (:owner %))
--- a/backend/test/api/base_test.clj
+++ b/backend/test/api/base_test.clj
@@ -10,6 +10,9 @@
 
 (defexpect game-creation
   (expect number? (:game-id (base/create-game)))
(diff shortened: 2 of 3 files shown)


backend/src/api/base.clj
```diff
@@ -51,6 +51,7 @@
                                   second-uuid]})
                   (assoc :game-id (:game-id lobby))
                   persistence/save-game
+                  :game-id
```
Masclins, line 54: So second player should get no game. But I assume due to the mocking errors was getting the game with :game-id 0?
Well, happy to see it's solved now.

backend/test/api/base_test.clj
```diff
@@ -9,15 +9,15 @@
 (ctest/use-fixtures :each mocking/mock-persistence)
 
 (defexpect game-creation
-  (expect 0 (:game-id (base/create-game)))
+  (expect number? (:game-id (base/create-game)))
```
Masclins, line 12: It would probably be interesting to see that creating *another* game, gives a different :game-id

backend/test/mocking.clj
```diff
@@ -1,15 +1,25 @@
 (ns mocking
```
Masclins, line 1: Sorry, I don't see myself reviewing this mocking. I don't get *how* it does what it does.

Masclins (changes requested):
Everything seems fine, but adding a new test may be worth it (see comments)

backend/test/mocking.clj
kenan-rhoton: `with-redefs` replaces function definitions within its context, and wherever we want it we apply it to all the `defexpect`s in a namespace.

The result is that instead of the real persistence functions being called, those defined here are called.

The other concept is simply that an `atom` is a *persistent* and *mutable* object which can be changed with `reset! atom value`.

If you have any other doubts I can clear them, I think it's important that I'm not the only one who understands this :upside_down_face:

backend/test/api/base_test.clj
kenan-rhoton: Agree!

backend/src/api/base.clj
kenan-rhoton: Mocking would always previously return the saved game on a fetch, regardless of the ID passed

kenan-rhoton:
A possibility for the future would be to split persistence into a set of functions that does what we want (generate IDs, save and fetch games) using a basic interface of `save` and `fetch` and have two implementations for that interface: a Redis DB based one and an In-Memory implementation, selectable at runtime.

This would eliminate mocking altogether as a necessity, as long as we select the In-Memory implementation for tests.

Might totally not be worth it though, considering how simple our database currently is (and seemingly will be for a while)

Masclins (approved):

Issue #131, opened 2018-08-31
Use a tool for cleaner code


Masclins:
I think it'd be a good idea to use  Codacy, Codefactor and/or a similar tool to ensure having a cleaner code and finding potential problems.

I use them both for signepedia and seems useful.

- [x] Clojure
- [x] Javascript / Node.js
- [ ] Other (pug, bash, Dockerfiles...)

kenan-rhoton:
You set up Codacy for this, right?

Masclins:
I did, but wasn't sure if everybody saw it or only me (since I'm not the admin of the project).

Also, can everybody see the Codacy reported issues? ( https://app.codacy.com/app/masclins/card-game/issues/index )

If so, this issue can be closed.

Masclins:
...I just realized that Codacy doesn't analyze Clojure (nor Leiningen)...

I'll look for another one and unset Codacy for this project.

Masclins:
It seems like github offers no tool for Clojure :(

Even though, there seems to exist Eastwood that can be used. It won't be integrated into github, but can be runned once in a while, or even be included in Docker or Travis, so it is somewhat automatized and we always have a file with the errors...

kenan-rhoton:
I can add `lein-cloverage` to Travis, I believe. I would make it a separate step so we're always clear on failure reason.

Masclins:
@kenan-rhoton if you include Codacy as a hard(ish) requirement to accept a PR, this Issue will be solved.

kenan-rhoton:
Closing due to Codacy and all the other actions taken so far.

PR #141, opened 2018-09-10
Cleaner code

### LINKED ISSUE #130: Use tool to track the Cover
We should begin using a tool that tracks the Cover of our tests. 

- [x] Backend tests coverage
- [ ] Frontend tests coverage

Masclins:
**Description**

Trying to have a better code I checked some code-reviewing tools for clojure. This PR is the result of applying their advices.

Also I added two of them that seemed the most useful.

**Issues Resolved**

This *might* solve #130 and #131, though we might want to include some other tools, auto-call them on Travis and/or keep track fo their reports through some other tool.

**Details**

Now Yagni and Cloverage are added to `project.clj` so they can be called.

I also changed the Issue Template for bug reports. Just the title for description of the bug.

**Hotspots**

`backend/src/api/create_player.clj` is really not used? I think it isn't...


kenan-rhoton (dismissed):
I can confirm `backend/src/api/create_player.clj` is an artifact from when we used to "half-create" a game in the API instead of just letting Rules do it when we have all the players already.

I also think it would be valuable to consider adding Yagni and Cloverage to the Travis test stack for `develop` and `master`

backend/src/rules/victory_conditions.clj
```diff
@@ -17,42 +17,31 @@
...
-        (let [tmp-wons (get-won-rows game-state (first player-ids))]
-          (cond (= tmp-wons most-wons)
-                (recur "" most-wons (rest player-ids))
-                (> tmp-wons most-wons)
-                (recur (first player-ids) tmp-wons (rest player-ids))
-                :else
-                (recur winner most-wons (rest player-ids))))))))
-
+    (cond
+      (nil? player-ids) ""
```
kenan-rhoton, line 40: This is to pass a certain test case where there are no player-ids... I think we should just consider that an invalid input and return an `:error` though, hence why I put a special case here.

I reordered the comparisons for readability.

kenan-rhoton (dismissed):

kenan-rhoton (approved):

PR #145, opened 2018-09-13
quotes, semicolons and arrowfunctions

Masclins:
**Description**

Following our recent changes for a clearer, cleaner and more consistent code, I solved some more style issues from Codacy

**Issues Resolved**

Many from Codacy regarding the use of double-quotes, semi-colons and parenthesis for arrow-functions.

**Details**

There are still many things to change, but I decided to stop here.
Files for e2e tests hadn't been changed since we'll probably need to redo them anyway.
Also, we have some issues regarding HTML injection. I'd prefere to wait to solve that, since @pernilsalat might change many of those for his issue.

### DIFF
--- a/frontend/src/config/config.js
+++ b/frontend/src/config/config.js
@@ -2,7 +2,7 @@ const config = require("./data/config.yml");
 
 // Load server data with defaults if file doesn't exist
 try {
-    const servers = require("./data/servers.yml");
+    var servers = require("./data/servers.yml");
 } catch(error) {
 
     // Don't want to silently catch other errors
@@ -11,7 +11,7 @@ try {
     }
 
     // Defaults to dev values
-    const servers = {
+    var servers = {
         servers: {
             frontend: "frontend:8880",
             backend: "backend:3000"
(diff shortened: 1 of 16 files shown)


frontend/src/config/config.js
```diff
@@ -11,7 +11,7 @@ try {
     }
 
     // Defaults to dev values
-    servers = {
+    const servers = {
```
kenan-rhoton: this breaks tests I think, we may be relying on supplanting this value in them

kenan-rhoton (dismissed):
Thanks for insisting on improving code quality!

frontend/src/config/config.js
```diff
@@ -2,7 +2,7 @@ const config = require("./data/config.yml");
 
 // Load server data with defaults if file doesn't exist
 try {
-    const servers = require("./data/servers.yml");
+    var servers = require("./data/servers.yml");
```
kenan-rhoton: The correct way I believe is to have a `var servers` declaration before the try and remove the `var` from everywhere else in the file

kenan-rhoton (approved):

PR #146, opened 2018-09-13
Invalid target

Masclins:
**Description**

I realized we weren't enforcing that the ability to add-power was applied to a card on the board.
This includes this rule in the backend.

**Issues Resolved**

None we've created *yet*.

**Details**

Just creates a new message and condition for `(play-card`.
It's worth noting that when we define and implement new card abilities, we'll probably rethink most of this function and how it handles targets.

**Hotspots**

I don't feel like any test is needed for the *api*, since we changed it to simply replicate *rules*' `:error`, but maybe somebody would like to add a test for it...?

### DIFF
--- a/backend/src/configs/messages.clj
+++ b/backend/src/configs/messages.clj
@@ -18,6 +18,8 @@
 
 (def need-target (get-in config-file [:messages :need-target]))
 
+(def invalid-target (get-in config-file [:messages :invalid-target]))
+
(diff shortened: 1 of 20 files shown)


kenan-rhoton (approved):
Api test is not needed since we've already validated that errors correctly propagate with other tests, and we should avoid duplication of code wherever possible (even in tests!)

PR #147, opened 2018-09-17
show card-pow when in row

Masclins:
**Description**

Resolves a Bug: Api was not sending the power of opponent's card already played.

(also adds a missed semicolon on another file I found recently)

**Issues Resolved**

Not created but reported on Discord by @pernilsalat 

**Details**

I messed up when creating the test, since I did not take into account that cards on rows should always show their power.

**Hotspots**

Should we create some other test?

### DIFF
--- a/backend/src/api/player_view_functions.clj
+++ b/backend/src/api/player_view_functions.clj
@@ -6,10 +6,18 @@
   "Returns cards as seen by a player"
   [game-state player-id]
   (vec (map
-         #(if (= (:owner %) player-id)
-            (assoc % :owner "me")
-            {:location (:location %)
-             :owner "opp"})
+         #(cond
+           (= (:owner %) player-id)
+           (assoc % :owner "me")
+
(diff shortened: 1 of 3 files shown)


backend/src/api/player_view_functions.clj
```diff
@@ -6,10 +6,18 @@
...
   (vec (map
-         #(if (= (:owner %) player-id)
-            (assoc % :owner "me")
-            {:location (:location %)
-             :owner "opp"})
+         #(cond
+           (= (:owner %) player-id)
+           (assoc % :owner "me")
+
+           (= (get-in % [:location 0]) :row)
```
Masclins, line 13: La idea, ara mateix, és que només es mostri el `:power` i cap altre atribut (tamoc n'hi ha cap altre, ara mateix).

Haurem de decidir què es mostra i què no quan creem noves habilitats, i modificar aquesta funció com correspongui.

kenan-rhoton (approved):

Issue #150, opened 2018-09-18
New ability: Card power modified depending on the row


Masclins:
We will include a new rule:

Cards might have different values on certain rows.

My idea is to give rows the attribute `:type "row's type"`. Some cards might have abilities that change their power depending on row's type.

Masclins:
I'm willing to solve this as follows:

- Rows can have *type*
- Cards can have an ability that changes its power when played on a row with a specific *type*.

It's explained on the [Manual](https://github.com/kenan-rhoton/card-game/wiki/0.-Troll:-Card-Game-Rules)

kenan-rhoton:
I agree with this approach

Masclins:
Trying to write this ability down I realized the function needs to know which card has the ability and on which row will be played.
On the other hand, `add-power` needed only the target.

To avoid having to pass different arguments to every ability, it seems the wisest would be always passing `game-state` and `play` (a single play, the one that triggered the ability)

kenan-rhoton:
that seems quite sensible, really, then it wouldn't even need a target

PR #152, opened 2018-09-19
Abilities as functions on backend

### LINKED ISSUE #149: Abilities as functions
After discussing about how to handle new card abilities we realized that Clojure gives us the opportunity to use functions as an attribute of cards.

We should make cards have (when appropiate) abilities as functions that are called when played.

That would give us a lot of flexibility on what cards can do.

Masclins:
**Description**

This changes how we understand and handle cards' abilities.
They'll be functions that are called when the card is played.

**Issues Resolved**

This resolves #149 and prepares everything to make #150 (and any further ability) easier to implement.

**Details**

There'd been no change on the response to the frontend. That means that frontend will recieve a stringified function which will be useless.

I did the PR anyway for the following reasons:
- Everything changed is tested and tests passes.
- Everything clearly stated on #149 is resolved.
- Which led me to: Probably we should make another Issue with it's own discussion for deciding how to handle the communication of abilities.

**Hotspots**

`rules.abilities/add-power` and `rules.play-card/apply-ability` should be **really** clear for all developers, since it's how abilities will work.


backend/src/rules/abilities.clj
```diff
@@ -0,0 +1,16 @@
...
+(defn add-power
+  "Creates the ability function"
+  [increase]
+  (fn    
+;    "Alters a cards' power, by adding the a defined value to it.
+;    This power may be negative, resulting in a decrease"
+    [game-state target]
+      (update-in
+        game-state
+        [:cards target]
```
kenan-rhoton: Wouldn't this be clearer and simpler?

```clojure
(update-in game-state [:cards target :power] #(+ % increase))
```

backend/src/rules/play_card.clj
```diff
@@ -1,37 +1,32 @@
 (ns rules.play-card
   (:require [configs.messages :as messages]
-            [rules.alter-card :as alter-card]
             [rules.count-cards :as count-cards]))
 
 (defn move-card
```
kenan-rhoton: this would be clearer as `move-card-to-row` or leaving it as is but receiving a location instead of a `row-id`

backend/src/rules/play_card.clj
```diff
@@ -1,37 +1,32 @@
...
-  [game-state play]
-  (let [add-power (get-in game-state
-                          [:cards (:card-id play) :add-power])
-        target (:target play)]
-    (if (nil? add-power)
-        game-state
-        (alter-card/add-power game-state target add-power))))
+(defn apply-ability
+  "Apply the ability of a play, if it exists"
+  [game-state next-play]
```
kenan-rhoton: I don't like the name `next-play` here, since we don't care about that. It's just the data indicating a play, so simply `play` seems better.

backend/src/rules/play_card.clj
```diff
@@ -1,37 +1,32 @@
...
 
-(defn apply-play
-  "Applies an entire play"
-  [game-state play]
-  (-> game-state
-      (apply-add-power play)
-      (move-card (:card-id play) (:row-id play))))
-
-(defn apply-all-plays
+(defn ^:private apply-all-plays
```
kenan-rhoton, line 18: Thank you for marking it private 🙃

kenan-rhoton (dismissed):
Amazing! Just some minor changes I think on this one

backend/src/rules/play_card.clj
```diff
@@ -1,37 +1,32 @@
 (ns rules.play-card
   (:require [configs.messages :as messages]
-            [rules.alter-card :as alter-card]
             [rules.count-cards :as count-cards]))
 
-(defn move-card
+(defn move-card-to-row
```
kenan-rhoton, line 5: I think I actually prefer this option:

```clojure
(defn move-card
  "Moves a card to the specified location"
  [game-state card-id location]
  (assoc-in game-state [:cards card-id :location] location))
```

Which would be used like:

```clojure
(move-card (:card-id play) [:row (:row-id play)])
```

What was the reason for choosing this option instead? I can be convinced 🙃

Masclins: `(move-card` will only be used to move cards to row.

And
`(move-card-to-row game-state (:card-id play) (:row-id play))` is more readable than
`(move-card game-state (:card-id play) [:row (:row-id play)])`

Still, I think we can do something in between. I create the function you perefer (and test it) and
```clojure
(defn ^:private move-card-to-row
  "Moves a card to the specified row"
  [game-state card-id row-id]
  (move-card game-state card-id [:row row-id]))
```

kenan-rhoton (approved):
Approved!

backend/src/rules/play_card.clj
kenan-rhoton: I'm convinced 😉

Issue #120, opened 2018-04-25
The Server needs to be available


Masclins:
We need http://card-game.rhoton.es/ up again.

Even though there are still many (important) things to be solved, soon (and by that I mean as soon as #119 is cleaned and merged) we'll be able to begin playtesting.

For our designing philosophy (doing so iteratively) we'll need **a lot** of playtesting. And it seems we'll have really few playtesters. Therefore, the sooner we begin with that, the better. And we can't do that without the card-game.rhoton.

kenan-rhoton:
This is being... problematic. We might need to consider alternate solutions (such as hosting it somewhere for a small fee)

Masclins:
Soon I'll need a host too for ensignapedia's project. Maybe we can host both there (while both will have next to 0 traffic)?

kenan-rhoton:
Ok, starting on this one!

kenan-rhoton:
Apparently the server is not available? @ruchnar

kenan-rhoton:
Have been talking about it with @ruchnar and we leave this on hold until he's back

kenan-rhoton:
We're going to try for a serverless architecture...

The two options seem to be:

1. AWS Lambda + AWS Elasticache (Redis)
2. Google Cloud Functions using ClojureScript compilation + Cloud Redis

I will try them in this order (since using ClojureScript seems like an unnecessary complication that could be avoided)

kenan-rhoton:
Mini-update:

AWS Lambda is looking cool. This will require a rewrite of the handler and persistence, but luckily we've made it so those files are essentially just protocol translations so it should be pretty painless.

The part that really worries me: I'll have to think of a way to make it also usable locally for development... Expect a pull request within 48 hours with a bunch of hotspots I'll need people to look at.

PR #156, opened 2018-09-20
Backend Deployed to AWS

kenan-rhoton:
**Description**

These changes are intended to smoothly integrate the backend into a AWS structure

**Issues Resolved**

This relates to #120, solving the Backend aspect of it.

**Details**

So there are quite a few changes:

1. Eliminated compojure to use an AWS Lambda-Compliant method, where `-handleRequest` is the entrypoint.
2. Switched from Redis to DynamoDB (ran into many problems using Amazon's Redis...)
3. Configuration is now hard-coded since the Lambda function doesn't have access to the YAML

**Hotspots**

The entire API Handler has been changed. However, one of the side-effects is that it is almost entirely testable now (which means I should definitely add tests...), except for the `-handleRequest` itself.

It should still serve the same exact API, although right now I haven't figured out how to run it locally to test it (I have been doing executions using the REPL to test it locally so far...)

This PR should still be considered Work In Progress


kenan-rhoton:
I think this PR needs to be blocked until frontend correctly matches with the current backend, maybe?

I'm afraid many frontend changes will have to be done to get everything working properly... Or maybe not, I'll think about it

PR #158, opened 2018-09-21
Frontend Coverage

Masclins:
**Description**

Coverage of the frontend is done and send to Codacy.

**Issues Resolved**

#130 

**Details**

To work correctly, @kenan-rhoton will need to add the environment variable CODACY_PROJECT_TOKEN
We might also add some variables to decide if some files must be taken into account even though they aren't tested

**Hotspots**

Which files should be considered in the coverage even if there is no test for them?


frontend/package.json
```diff
@@ -2,6 +2,7 @@
   "license": "MIT",
   "dependencies": {
     "clean-webpack-plugin": "^0.1.19",
+    "codacy-coverage": "^3.1.0",
```
kenan-rhoton: this should be in "devDependencies" (unneeded in production)

kenan-rhoton (changes requested):

kenan-rhoton (approved):

PR #162, opened 2018-09-24
Abilities to fe

### LINKED ISSUE #154: Backend passes abilities info to frontend
**Description of the current (after #152) situation**

Card abilities are functions. That makes it impossible for frontend to understand what the ability does and therefore what it should show or ask the player.

**Description of the wanted situation**

Abilities functions can be called with no argument. That returns a definition (map?) that can be handled by fronted and gives information about what to show and what to expect.

For *add-power +3* ability I was thinking about something like that:
For backend, card would have: `:ability (ability/add-power 3)`.
For frontend, card would have: `:ability {:add-power 3 :targets 1}`. That gives information on the ability and it's *related attribute*, as well as how many targets it should expect.  

**Considered alternatives**

I think we should pass the number of needed targets too. Maybe we could make it somehow different, like `:ability {:add-power 3} :needed-targets 1`.

Options that put all the ability on a string or let the frontend decide if the player must choose a target seem too much work for the frontend (extreme case would be `:ability "add-power-3"`)

**Additional context**

It's probably impossible to choose the *best* option when we don't know yet what kind of abilities we'll have, but I'd want to give the most importance to:

- Keeping it the most flexible.
- Avoiding as much work from frontend as possible.

We won't be able to make any e2e advance if we don't solve this. If and when we merge #152 e2e communication will be broken until we solve it.

Masclins:
**Description**

Backend responding API requests only gives relevant information about cards.

**Issues Resolved**

Resolves #154 

**Details**

All cards abilities (which are functions) must be created in a way that calling them without parameters gives the frontend relevant information. That's a `:description` and `:target`, as proposed by @kenan-rhoton on #154.

**Hotspots**

`(api.player_view_functions/get-cards` should be carefully reviewed. I tried to make it the most readable by dividing the 4 possible scenarios regarding card's info that must be shown and write every time what should be send. An alternative would be abuse `(merge` to avoid repetitions, but I felt that would be less redeable.

(I leave the checks for reviewers)

- [ ] The tests pass
- [ ] The code has tests
- [ ] The tests are well-designed
- [ ] The code is readable
- [ ] The code is well-organized

### DIFF
--- a/backend/src/api/player_view_functions.clj
+++ b/backend/src/api/player_view_functions.clj
@@ -6,18 +6,27 @@
   "Returns cards as seen by a player"
   [game-state player-id]
   (vec (map
-         #(cond
-           (= (:owner %) player-id)
-           (assoc % :owner "me")
+         #(if (= (:owner %) player-id)
 
-           (= (get-in % [:location 0]) :row)
-           {:power (:power %)
-            :location (:location %)
-            :owner "opp"}
+            (if (and (contains? % :ability) (= (get-in % [:location 0]) :row))
+              (merge 
+                {:power (:power %)
+                 :location (:location %)
+                 :owner "me"}
+                ((:ability %)))
 
-           :else
-           {:location (:location %)
-            :owner "opp"})
+              {:power (:power %)
+               :location (:location %)
+               :owner "me"})
+
+            (if (= (get-in % [:location 0]) :row)
+              {:power (:power %)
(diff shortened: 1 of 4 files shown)


Masclins:
@kenan-rhoton Codacy's token's missing :( (That's why Travis fails)

Masclins:
Close to reopen from Kenan's project

PR #164, opened 2018-09-25
Row types

### LINKED ISSUE #150: New ability: Card power modified depending on the row

Masclins:
**Description**

New ability created: Now cards can have an ability that makes their power change when played on specific rows.

Rows can have a `:type` and this ability checks if the card had been played on a row of the appropriate type.

Also, I changed `(add-power` so all abilities recieve `game-state` and `play`.

**Issues Resolved**

Resolves #150 

**Hotspots**

`(rules.abilities/type-add-power` is probably what's most relevant.

Still, I find it **very important** to double-check that we have all the tests we'd want to.

- [x] The tests pass
- [x] The code has tests
- [x] The tests are well-designed
- [ ] The code is readable
- [ ] The code is well-organized
- [x] Any rule changed or included is also on the Manual

### DIFF
--- a/backend/src/rules/abilities.clj
+++ b/backend/src/rules/abilities.clj
@@ -12,4 +12,7 @@
        [:cards target :power]
        #(+ % increase)))
      ([]
-      {:description (str "Enhance " increase) :target 1})))
+      {:description (if (neg? increase)
+                      (str "Weaken " (- increase))
+                      (str "Enhance " increase))
+       :target 1})))
--- a/backend/src/rules/abilities.clj
+++ b/backend/src/rules/abilities.clj
@@ -3,28 +3,29 @@
 (defn add-power
   "Creates the ability function"
   [increase]
-  (fn    
-;   "Alters a cards' power, by adding the a defined value to it.
-;   This power may be negative, resulting in a decrease"
-    ([game-state play]
+  {:ability
+   (fn    
+;    "Alters a cards' power, by adding the a defined value to it.
+;    This power may be negative, resulting in a decrease"
+     [game-state play]
      (update-in
        game-state
        [:cards (:target play) :power]
        #(+ % increase)))
-     ([]
-      {:description (if (neg? increase)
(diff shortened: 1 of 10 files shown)


Masclins:
Errors on tests now are because it's implemented that all abilities require a target.
I'll change that when #165 is merged, since I'll use what's implemented there (that returns the number of needed targets)

backend/src/rules/abilities.clj
```diff
@@ -6,13 +6,34 @@
...
+    ([game-state play]
+     (if (= (get-in game-state [:rows (:row-id play) :type])
+            row-type)
+       (update-in
+         game-state
+         [:cards (:card-id play) :power]
+         #(+ % increase))
+
+       game-state))
+    ([]
```
kenan-rhoton: I have the enormous suspicion that calling the function with no arguments to get a description is a bad idea... but I can't think of a better alternative.

Maybe to make it more reasonable we can have a function `(describe ability)` which simply calls `ability` with no parameters? I don't know, something seems very off to me.

kenan-rhoton: Oh wait, I have an idea. The function should generate both a function and a description everytime, so an ability has essentially `{:function (fn ...) :description "This is the description")`.

This separates the concerns cleanly instead of having a function return radically different things dependent on the number of arguments.

kenan-rhoton (dismissed):

backend/src/rules/abilities.clj
Masclins: I really like it.
Therefore cards would have `:ability some-function :description "ability description"` and, when needed, `:target 1`.

Is that right?
I'll do that change when able.

kenan-rhoton: Thanks!

PR #165, opened 2018-09-25
Abilities to fe

### LINKED ISSUE #154: Backend passes abilities info to frontend

Masclins:
**Description**

Backend responding API requests only gives relevant information about cards.

**Issues Resolved**

Resolves #154 

**Details**

All cards abilities (which are functions) must be created in a way that calling them without parameters gives the frontend relevant information. That's a `:description` and `:target`, as proposed by @kenan-rhoton on #154.

**Hotspots**

`(api.player_view_functions/get-cards` should be carefully reviewed. I tried to make it the most readable by dividing the 4 possible scenarios regarding card's info that must be shown and write every time what should be send. An alternative would be abuse `(merge` to avoid repetitions, but I felt that would be less redeable.

(I leave the checks for reviewers)

- [x] The tests pass
- [x] The code has tests
- [x] The tests are well-designed
- [x] The code is readable
- [x] The code is well-organized

### DIFF
--- a/backend/src/api/player_view_functions.clj
+++ b/backend/src/api/player_view_functions.clj
@@ -6,18 +6,27 @@
   "Returns cards as seen by a player"
   [game-state player-id]
   (vec (map
-         #(cond
-           (= (:owner %) player-id)
-           (assoc % :owner "me")
+         #(if (= (:owner %) player-id)
 
-           (= (get-in % [:location 0]) :row)
-           {:power (:power %)
-            :location (:location %)
-            :owner "opp"}
+            (if (and (contains? % :ability) (= (get-in % [:location 0]) :row))
+              (merge 
+                {:power (:power %)
+                 :location (:location %)
+                 :owner "me"}
+                ((:ability %)))
 
-           :else
-           {:location (:location %)
-            :owner "opp"})
+              {:power (:power %)
+               :location (:location %)
+               :owner "me"})
+
+            (if (= (get-in % [:location 0]) :row)
+              {:power (:power %)
+               :location (:location %)
+               :owner "opp"}
+
+              {:location (:location %)
+               :owner "opp"}))
+
          (:cards game-state))))
 
 (defn get-rows
--- a/backend/src/api/player_view_functions.clj
+++ b/backend/src/api/player_view_functions.clj
@@ -8,7 +8,7 @@
   (vec (map
          #(if (= (:owner %) player-id)
 
-            (if (and (contains? % :ability) (= (get-in % [:location 0]) :row))
+            (if (and (contains? % :ability) (= (get-in % [:location 0]) :hand))
               (merge 
                 {:power (:power %)
                  :location (:location %)
--- a/backend/src/api/player_view_functions.clj
+++ b/backend/src/api/player_view_functions.clj
@@ -2,31 +2,31 @@
   (:require [rules.victory-conditions :as victory]
             [configs.messages :as messages]))
 
+(defn ^:private partial-card
+  "Returns only certain keys of a card and change the :owner"
+  [card wanted-keys owner]
+  (assoc (select-keys card wanted-keys)
+         :owner owner))
+
 (defn get-cards
   "Returns cards as seen by a player"
   [game-state player-id]
   (vec (map
-         #(if (= (:owner %) player-id)
+         #(cond
+            (and (and (= (:owner %) player-id)
+                      (contains? % :ability))
+                      (= (get-in % [:location 0]) :hand))
+            (merge (partial-card % [:power :location] "me")
+                   ((:ability %)))
 
-            (if (and (contains? % :ability) (= (get-in % [:location 0]) :hand))
-              (merge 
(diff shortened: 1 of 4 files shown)


pernilsalat:
Ehhh, i don't think #154 discussion is finished...

backend/src/api/player_view_functions.clj
```diff
@@ -6,18 +6,27 @@
   "Returns cards as seen by a player"
   [game-state player-id]
   (vec (map
```
kenan-rhoton, line 14: This function is becoming very long and hard to read... do we really need nested ifs?
I feel like `cond` would be clearer, even if the conditions become a little more verbose, and maybe even split off the different cases into functions?

backend/src/rules/abilities.clj
```diff
@@ -4,10 +4,12 @@
...
-      #(+ % increase))))
+;   "Alters a cards' power, by adding the a defined value to it.
+;   This power may be negative, resulting in a decrease"
+    ([game-state target]
+     (update-in
+       game-state
+       [:cards target :power]
+       #(+ % increase)))
+     ([]
+      {:description (str "Enhance " increase) :target 1})))
```
kenan-rhoton: "Enhance" doesn't seem very appropriate for a negative value

backend/test/api/player_view_functions_test.clj
```diff
@@ -9,13 +10,27 @@
...
-                                  {:power 1 :attr "kill" :location [:hand] :owner "me_name"}
+                                  {:power 1 :internal-attr "kill" :location [:hand] :owner "me_name"}
                                   {:power 100 :location [:row 1] :owner "me_name"}]}
-                        "me_name")))
+                        "me_name"))
+
+  ; Correctly returns cards with abilities
+  (expect
+    [{:power 1 :location [:hand] :owner "me" :description "Enhance 100" :target 1}
+     {:power 17 :location [:hand] :owner "me" :description "Enhance -1" :target 1}
```
kenan-rhoton: Again, I feel like "Enhance -1" just looks very weird. Diminish or Weaken maybe? It might even then be positive "Weaken 1"

kenan-rhoton (dismissed):
In general good, but some readability issues and I don't like using "Enhance" for negative values"

backend/src/api/player_view_functions.clj
Masclins: Let's see if it's easier to read now

backend/src/api/player_view_functions.clj
```diff
@@ -2,31 +2,31 @@
   (:require [rules.victory-conditions :as victory]
             [configs.messages :as messages]))
 
+(defn ^:private partial-card
+  "Returns only certain keys of a card and change the :owner"
```
Masclins, line 6: I know that this could be a more generic function (no need to call it "card") but since it changes the `:owner`, I felt like it should be like that.

Also, considering it's private, I see no reason to make it more generic.

kenan-rhoton: agree with no need to be more generic

backend/src/api/player_view_functions.clj
```diff
@@ -2,22 +2,31 @@
...
+         :owner owner))
+
 (defn get-cards
   "Returns cards as seen by a player"
   [game-state player-id]
   (vec (map
          #(cond
-           (= (:owner %) player-id)
-           (assoc % :owner "me")
+            (and (and (= (:owner %) player-id)
```
kenan-rhoton: double `and` is redundant. In clojure `and` takes any amount of arguments, so you can just do `(and cond1 cond2 cond3 ... condN)`

kenan-rhoton (changes requested):
Awesome! Just a small "double and" issue (which is a Clojure syntax thing)

kenan-rhoton (approved):

